PR #34, opened 2016-03-02
Add YouCompleteMe configuration

conradev:
It uses ninja to determine the compile flags for a given file

Not sure if this has a place in the repo, but I find it convenient when editing the project in `vim`!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 # Build
 /build/
 
+# YCM
+.ycm_extra_conf.pyc
+
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -1,0 +1,55 @@
+import os
+import subprocess
+
+
+def FlagsForFile(filename):
+  flags = []
+
+  root_dir = os.path.normpath(os.path.dirname(filename))
+  while not (os.path.exists(os.path.join(root_dir, 'Sources'))):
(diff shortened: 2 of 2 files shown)


grp:
Interesting! It looks like some of the code for this might come from Chromium? (https://chromium.googlesource.com/chromium/chromium/+/trunk/tools/vim/chromium.ycm_extra_conf.py) If so, probably need to add it to the LICENSE document or at least the file header. (I'll also need to check that it's OK, but it should be fine.)

It might be cool to make a tool in xcbuild that makes YouCompleteMe work with any file in an Xcode project. It should be pretty easy: if you know the project file, just load that project and generate a clang invocation for just the file you need. It might be best implemented as a server so it doesn't have to regenerate the target build settings every time and can easily do a reverse lookup of the file -> project mapping.

conradev:
Yep, can do that, where should I add it?

Re: the second idea, that is something I am drooling to make. It would honestly probably be more reliable than Xcode's highlighting, which goes on and off, has phantom errors/warnings, etc. YCM has its own daemon (`ycmd`), I wonder if it would be possible to integrate into that.

b3ll:
:100:

grp:
Hi, unfortunately we cannot accept this pull request. You can put this configuration into a separate repository to be used alongside xcbuild.

PR #49, opened 2016-05-23
Build with GCC 4.8/4.9

Ktwu:
I'm working on building xcbuild on some centOS machines with GCC 4.8/4.9 installed.  Upgrading GCC to 5.0 isn't an option since I don't control the packaging available, so I opted to make xcbuild happily build instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 # Enable all warnings.
 add_compile_options(-Wall -Werror)
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-  add_compile_options(-Wno-multichar -Wno-sign-compare)
+  add_compile_options(-Wno-multichar -Wno-sign-compare -fno-strict-aliasing)
 endif ()
 
 # Enable color diagnostics.
--- a/Libraries/libutil/Headers/libutil/Subprocess.h
+++ b/Libraries/libutil/Headers/libutil/Subprocess.h
@@ -33,7 +33,9 @@ public:
                  std::ostream *output = nullptr,
                  std::ostream *error = nullptr)
     {
-        return execute(path, { }, { }, "", input, output, error);
+        std::vector<std::string> emptyVector;
+        std::unordered_map<std::string, std::string> emptyMap;
+        return execute(path, emptyVector, emptyMap, "", input, output, error);
     }
 
     bool execute(std::string const &path,
@@ -42,7 +44,8 @@ public:
                  std::ostream *output = nullptr,
                  std::ostream *error = nullptr)
     {
-        return execute(path, arguments, { }, "", input, output, error);
+        std::unordered_map<std::string, std::string> emptyMap;
+        return execute(path, arguments, emptyMap, "", input, output, error);
     }
 
     bool execute(std::string const &path,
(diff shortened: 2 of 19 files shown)


CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 # Enable all warnings.
 add_compile_options(-Wall -Werror)
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-  add_compile_options(-Wno-multichar -Wno-sign-compare)
+  add_compile_options(-Wno-multichar -Wno-sign-compare -fno-strict-aliasing)
```
grp: Any reason this is needed? I suppose it would be nicer to leave it on if possible. Did it cause test failures?

Libraries/libutil/Headers/libutil/Subprocess.h
```diff
@@ -33,7 +33,9 @@ class Subprocess {
                  std::ostream *output = nullptr,
                  std::ostream *error = nullptr)
     {
-        return execute(path, { }, { }, "", input, output, error);
+        std::vector<std::string> emptyVector;
+        std::unordered_map<std::string, std::string> emptyMap;
```
grp: Unfortunate, but old compilers are old compilers I suppose.

Libraries/xcsdk/CMakeLists.txt
```diff
@@ -23,3 +23,4 @@ target_include_directories(xcsdk PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Headers")
 add_executable(xcrun Tools/xcrun.cpp)
 target_link_libraries(xcrun xcsdk util)
 
+install(TARGETS xcrun RUNTIME DESTINATION bin)
```
grp: Could you move this to a separate diff? Might be good to add `install()` calls for all the utilities together.

I also wonder if this would need `install()` calls for the shared libraries too? I'm not sure how CMake handles that with install names on OS X.

Libraries/pbxbuild/CMakeLists.txt
```diff
@@ -69,6 +69,6 @@ target_link_libraries(dump_hmap pbxbuild util plist)
 
 ADD_UNIT_GTEST(pbxbuild DirectedGraph Tests/test_DirectedGraph.cpp)
 ADD_UNIT_GTEST(pbxbuild OptionsResolver Tests/test_OptionsResolver.cpp)
-target_link_libraries(test_pbxbuild_OptionsResolver PRIVATE pbxspec pbxsetting plist)
+#target_link_libraries(test_pbxbuild_OptionsResolver PRIVATE pbxspec pbxsetting plist)
```
grp: Any reason you disabled this?

CMakeLists.txt
Ktwu: It caused build failures.  There are some parts that would try to cast the address of doubles as the address of uint64_t's, which caused this to trip.  __ABPWriteDate() is one such place this happened.

Libraries/libutil/Headers/libutil/Subprocess.h
Ktwu: Not really.  The only advantage of this is that the line using these variables is shorter.

Libraries/xcsdk/CMakeLists.txt
Ktwu: Sure.  I don't know which targets need install(), so your advice there would be appreciated.

Libraries/pbxbuild/CMakeLists.txt
Ktwu: Oops!  This is a bit of leftover hacking from trying to get everything building.  I'll re-enable it.

CMakeLists.txt
grp: Interesting! Probably worth fixing at some point, but this seems fine for now.

strager: I object to this change!

grp: Then fix it! ;P

PR #50, opened 2016-05-25
Support `make install` with `PREFIX=` to install xcbuild.

grp:
With this, using `make install PREFIX=/tmp`, the files output are:

    /tmp/lib/libacdriver.dylib
    /tmp/lib/libbom.dylib
    /tmp/lib/libbuiltin.dylib
    /tmp/lib/libcar.dylib
    /tmp/lib/libdependency.dylib
    /tmp/lib/libext.dylib
    /tmp/lib/libninja.dylib
    /tmp/lib/libpbxbuild.dylib
    /tmp/lib/libpbxproj.dylib
    /tmp/lib/libpbxsetting.dylib
    /tmp/lib/libpbxspec.dylib
    /tmp/lib/libplist.dylib
    /tmp/lib/libutil.dylib
    /tmp/lib/libxcassets.dylib
    /tmp/lib/libxcdriver.dylib
    /tmp/lib/libxcexecution.dylib
    /tmp/lib/libxcformatter.dylib
    /tmp/lib/libxcscheme.dylib
    /tmp/lib/libxcsdk.dylib
    /tmp/lib/libxcworkspace.dylib
    /tmp/bin/actool
    /tmp/bin/builtin-copy
    /tmp/bin/builtin-copyPlist
    /tmp/bin/builtin-copyStrings
    /tmp/bin/builtin-copyTiff
    /tmp/bin/builtin-embeddedBinaryValidationUtility
    /tmp/bin/builtin-infoPlistUtility
    /tmp/bin/builtin-lsRegisterURL
    /tmp/bin/builtin-productPackagingUtility
    /tmp/bin/builtin-validationUtility
    /tmp/bin/dependency-info-tool
    /tmp/bin/lsbom
    /tmp/bin/plutil
    /tmp/bin/xcbuild
    /tmp/bin/xcrun

Confirmed the install paths are set up correctly to run the binaries on OS X.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,21 +9,29 @@
 
 build := build
 project := project
+PREFIX := /
 
 cmake := cmake
 cmake_flags := 
+cmake_command := $(cmake) -H. $(cmake_flags)
 
 ninja := $(if $(shell which llbuild),llbuild ninja build,ninja)
 ninja_flags := $(if $(shell echo "$$NINJA_JOBS"),-j$(shell echo "$$NINJA_JOBS"),)
+ninja_command := $(ninja) $(ninja_flags)
 
 all:
 	mkdir -p $(build)
-	$(cmake) -B$(build) -H. -G Ninja $(cmake_flags)
-	$(ninja) -C $(build) $(ninja_flags)
+	$(cmake_command) -B$(build) -GNinja
+	$(ninja_command) -C $(build)
 
 project:
 	mkdir -p $(project)
-	$(cmake) -B$(project) -H. -G Xcode $(cmake_flags)
+	$(cmake_command) -B$(project) -GXcode
+
+install:
+	mkdir -p $(PREFIX)
+	$(cmake_command) -B$(build) -GNinja -DCMAKE_INSTALL_PREFIX:PATH=$(realpath $(PREFIX)) -DCMAKE_INSTALL_NAME_DIR:STRING=@executable_path/../lib
+	$(ninja_command) -C $(build) install
 
 test: all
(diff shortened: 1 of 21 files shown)


Makefile
```diff
@@ -9,21 +9,29 @@
...
+	$(ninja_command) -C $(build)
 
 project:
 	mkdir -p $(project)
-	$(cmake) -B$(project) -H. -G Xcode $(cmake_flags)
+	$(cmake_command) -B$(project) -GXcode
+
+install:
+	mkdir -p $(PREFIX)
+	$(cmake_command) -B$(build) -GNinja -DCMAKE_INSTALL_PREFIX:PATH=$(realpath $(PREFIX)) -DCMAKE_INSTALL_NAME_DIR:STRING=@executable_path/../lib
```
strager: Why are we overriding CMAKE_INSTALL_PREFIX instead of using CMake's default? / is an annoying default; it'll install into /bin and /lib, which I doubt anyone expects.

Doesn't CMAKE_INSTALL_NAME_DIR affect compilation? I don't think the install target triggers compilation.

strager:
I don't like the Makefile wrapper. =<

All the CMake bits lgtm.

Makefile
grp: This does trigger compilation. The issue is that if you want to run it *without* installing, you need `CMAKE_INSTALL_NAME_DIR` to be set to `@rpath` (as per default), or else things won't run from the `build/` directory. So we have to set it and rebuild here.

How would you set the install prefix if `CMAKE_INSTALL_PREFIX` wasn't set here? `make all` doesn't pass anything to CMake.

PR #51, opened 2016-05-25
Car writer

marcsalem:
In this pull request:
* optimizations to car::Reader to speed up loading loading CAR files
* fixes bugs in libbom when opening a new file for writing
* added car::Writer and unit tests to libcar to write imagesets to CAR files


Libraries/libcar/Sources/AttributeList.cpp
```diff
@@ -99,3 +99,24 @@ operator()(AttributeList const &attributes) const
     /* Not a great hash. */
     return attributes.count();
 }
+
+
+std::vector<uint8_t> AttributeList::
+Write(size_t count,
+    uint32_t *identifiers) const
```
grp: nit: this could probably go on one line

Libraries/libcar/Sources/AttributeList.cpp
```diff
@@ -99,3 +99,24 @@ operator()(AttributeList const &attributes) const
...
+{
+    std::vector<uint8_t> output = std::vector<uint8_t>(sizeof(uint16_t)*count);
+    uint16_t *values = (uint16_t *) &output[0];
```
grp: nit: use `reinterpret_cast<uint16_t *>`. you can also use `output.data()` since C++ 11 instead of `&output[0]`

Libraries/libcar/Tests/test_AttributeList.cpp
```diff
@@ -0,0 +1,89 @@
...
+		car_attribute_identifier_dimension1
+	}
+};
+
+static struct car_key_format *keyfmt = &keyfmt_s.keyfmt;
+
+
+TEST(AttributeList, TestAttributeListDeSerialize)
+{
+	uint16_t rendition_key[13] = {
```
grp: nit: re-indent this block.

Libraries/libcar/Headers/car/Reader.h
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 #include <unordered_map>
+#import <experimental/string_view>
```
grp: I'm not sure this is going to work with GNU libstdc++. We might need to import the LLVM libc++ version of this into `ext/`, like is done with `ext/optional`.

Libraries/libcar/Headers/car/Facet.h
```diff
@@ -59,6 +59,14 @@ class Facet {
 
 public:
     /*
+     * Load an existing rendition matching the provided attributes.
+     */
+    static Facet Load(
```
grp: Would this commit be better as a separate PR? It doesn't look too related to writing. (But maybe you use it later, haven't gotten there yet.)

Libraries/libcar/Sources/Facet.cpp
```diff
@@ -69,3 +69,19 @@ Create(std::string const &name, AttributeList const &attributes)
     return Facet(name, attributes);
 }
 
+std::vector<uint8_t> Facet::Write() const
+{
+    size_t attributes_count = _attributes.count();
+    size_t facet_value_size = sizeof(struct car_facet_value) + (sizeof(struct car_attribute_pair) * attributes_count);
+    std::vector<uint8_t> output = std::vector<uint8_t>(facet_value_size);
+    struct car_facet_value *facet_value = (struct car_facet_value *)&output[0];
```
grp: nit: `reinterpret_cast` here too

Libraries/libcar/Tests/test_Facet.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto facet = car::Facet::Load(std::string("somename"), test_facet_value);
+    EXPECT_TRUE(facet.name() == std::string("somename"));
+    EXPECT_TRUE(*(facet.attributes().get(car_attribute_identifier_element)) == 85);
+    EXPECT_TRUE(*(facet.attributes().get(car_attribute_identifier_part)) == 181);
+    EXPECT_TRUE(*(facet.attributes().get(car_attribute_identifier_identifier)) == 4258);
+}
+
+TEST(Facet, TestFacetSerialize)
+{
+    std::vector<uint8_t> test_facet_value_vector((uint8_t*)test_facet_value,
```
grp: nit: reinterpret_cast as well

Libraries/libcar/Headers/car/Rendition.h
```diff
@@ -72,20 +72,49 @@ class Rendition {
...
 
 private:
     std::string _fileName;
     int         _width;
     int         _height;
     float       _scale;
+    bool        _is_vector;
+    bool        _is_opaque;
+    bool        _is_resizable;
+    enum resize_mode _resize_mode;
```
grp: nit: camelCase for these properties

Libraries/libcar/Headers/car/Rendition.h
```diff
@@ -72,20 +72,49 @@ class Rendition {
         { return _format; }
     };
 
+public:
+    enum resize_mode {
```
grp: Are these resizing modes a property of the car format? If so, might be better for `car_format.h`.

nit: you could use `enum class` here to use namespacing instead of prefixing for the values.

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -235,3 +507,139 @@ Create(
...
+
+std::vector<uint8_t> Rendition::Write()
+{
+    // Create header
+    struct car_rendition_value header;
+    bzero(&header, sizeof(struct car_rendition_value));
+    header.magic[3] = 'C';
+    header.magic[2] = 'T';
+    header.magic[1] = 'S';
+    header.magic[0] = 'I';
```
grp: nit: strncpy()?

Libraries/libcar/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ add_library(car SHARED
             Sources/Facet.cpp
             Sources/Rendition.cpp
             Sources/car_format.c
+            Sources/Writer.cpp
```
grp: nit: move next to `Reader.cpp`

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -0,0 +1,115 @@
...
+public:
+    /*
+     * Add a Facet
+     */
+    void addFacet(Facet const &);
+
+    /*
+     * Add a Rendition for a Facet, allow lazy loading of data
+     */
+    void addRendition(Rendition &);
```
grp: Any reason this isn't a const ref?

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -0,0 +1,115 @@
...
+public:
+    /*
+     * Lookup a Facet by name
+     */
+    ext::optional<car::Facet> lookupFacet(std::string name) const;
+
+    /*
+     * Lookup Rendition list for a Facet
+     */
+    std::vector<car::Rendition> lookupRenditions(Facet const &) const;
```
grp: Do you need to lookup and iterate functions on the writer? Seems like the common case would be just to add a bunch then write it out.

Libraries/libcar/Sources/Writer.cpp
```diff
@@ -0,0 +1,134 @@
...
+                (void *)item.first.c_str(), item.first.size(),
+                (void*)&facet_value[0], facet_value.size());
+        }
+        bom_tree_free(facets_tree_context);
+    }
+
+    // Write RENDITIONS
+    struct bom_tree_context *renditions_tree_context = bom_tree_alloc_empty(_bom.get(), car_renditions_variable);
+    if (renditions_tree_context != NULL) {
+       for ( auto& item : _renditionValues ) {
```
grp: style nit: for all of these for loops, the general style is `for (auto const &item : values)`

Libraries/libcar/Tools/dump_car.cpp
```diff
@@ -239,23 +239,22 @@ main(int argc, char **argv)
...
+    int rendition_count = 0;
+
+    car->facetIterate([&car, &facet_count, &rendition_count, output](car::Facet const &facet) {
+        facet_count++;
         facet.dump();
 
-        int rendition_count = 0;
-        facet.renditionIterate(&*car, [&rendition_count](car::Rendition const &rendition) {
+        auto renditions = car->lookupRenditions(facet);
+        for (auto & rendition : renditions) {
```
grp: nit: `auto const &rendition`

Libraries/libcar/Tests/test_Writer.cpp
```diff
@@ -63,9 +63,24 @@ static test_car_key_format keyfmt_s = {
...
 
-TEST(Writer, TestWriter)
+class WriterTest : public testing::Test {
+
+protected:
+    std::string tmpfilename;
+
+    virtual void SetUp() {
+        std::ostringstream stream;
+        stream << "/tmp/test_car_Writer_output." <<  ::getpid() << ".car";
```
grp: Instead of a temp file, could this use the memory-backed BOM source?

Libraries/libcar/Sources/AttributeList.cpp
marcsalem: Hard to see from this diff, but I was sticking with the style that was there, but I agree, it can go on one line. I'll clean it up.

Libraries/libcar/Headers/car/Facet.h
marcsalem: It's related to faster load time for the reader

Libraries/libcar/Headers/car/Writer.h
marcsalem: Ah, this is from experimenting with making the data/decode method save the result, which made most of Rendition references non-const, but after thinking about memory usage, I switched back. I'll clean this up.

Libraries/libcar/Headers/car/Writer.h
marcsalem: This was a last minute addition, I'll remove it.

Libraries/libcar/Tests/test_Writer.cpp
marcsalem: I guess it could, but this works pretty well, and it does also test the file backed mmap path which had bugs.

Libraries/libcar/Headers/car/Rendition.h
marcsalem: I actually find it strange that a person needs to use car_format.h to use libcar. I think over time libcar should replace all those enums and types with it's own, and keep car_format.h as internal header that describes just the file format, and not the C++ API to read and write the data.

grp: Agree in general. But I was thinking that would probably be a more C++ style API — this looked more like C so I wasn't sure which it was.

marcsalem: Yeah, that reasonable. I've already switched to enum class and I'm going through Linux build issues right now.

PR #52, opened 2016-05-25
Build with GCC 4.9 on Centos7

Ktwu:
With these changes, I'm able to build xcbuild with my setup and,
perhaps importantly, run one of the tools that uses libxml.


grp:
lgtm!

PR #57, opened 2016-05-26
Fix compilation with GCC 6 on Ubuntu

strager:
Building GCC 6.1.1 on Ubuntu 14.04  detected two errors:

* The return values of some APIs were not checked, and
* The result of some variables were not used.

Fix these errors, allowing GCC 6 to build all of xcbuild.

### DIFF
--- a/Libraries/libutil/Sources/SysUtil.cpp
+++ b/Libraries/libutil/Sources/SysUtil.cpp
@@ -52,12 +52,7 @@ EnvironmentVariables()
 }
 
 #if defined(__linux__)
-static char initialWorkingDirectory[PATH_MAX] = { 0 };
-__attribute__((constructor))
-static void InitializeInitialWorkingDirectory()
-{
-    getcwd(initialWorkingDirectory, sizeof(initialWorkingDirectory));
-}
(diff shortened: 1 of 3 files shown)


strager:
The getcwd change is wrong. getcwd behaves this way on OS X, but not necessarily on Linux/glibc.

Libraries/libutil/Sources/SysUtil.cpp
```diff
@@ -52,12 +52,7 @@ EnvironmentVariables()
 }
 
 #if defined(__linux__)
-static char initialWorkingDirectory[PATH_MAX] = { 0 };
-__attribute__((constructor))
-static void InitializeInitialWorkingDirectory()
-{
-    getcwd(initialWorkingDirectory, sizeof(initialWorkingDirectory));
-}
+static std::string initialWorkingDirectory = FSUtil::GetCurrentDirectory();
```
grp: Could you avoid a `static` non-POD type?

strager: Should I replace it with a GNU constructor? It's basically the same, except more verbose and less portable.

grp: Feels a little more explicit, I guess.

PR #64, opened 2016-06-16
Actool fixes

marcsalem:
This pull request has some improvements for writing a CAR file.


Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -27,10 +55,175 @@ CompileAction::
...
+static void
+CompileChildren(ext::optional<car::Writer> &writer, std::string ns, std::shared_ptr<xcassets::Asset::Asset> const &parent, std::vector<T> const &assets, Result *result)
+{
+    if (assets.empty()) {
+        return;
+    }
+
+    for (auto const &asset : assets) {
+        if (CompileContents(writer, ns, parent, asset, result)) {
+
```
Ktwu: What's with the empty if() block here?

Libraries/acdriver/Headers/acdriver/Options.h
```diff
@@ -43,7 +44,7 @@ class Options {
 private:
     std::string _platform;
     std::string _minimumDeploymentTarget;
-    std::string _targetDevice;
+    std::set<std::string> _targetDevice;
```
grp: nit: I'd use `std::vector` to preserve ordering here, it's more consistent with other options structures and is useful to avoid losing information from the options parsing process. Alternatively, could use `std::unordered_set`, since I doubt this needs to be sorted.

Libraries/xcassets/Headers/xcassets/Slot/ImageSize.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <ext/optional>
+#include <vector>
+#include <string>
+
+namespace xcassets {
+namespace Slot {
+
+/*
+ * The density ratio of a target device.
```
grp: nit: Looks like this is left over from another file.

Libraries/xcassets/Headers/xcassets/Slot/ImageSize.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+class ImageSize {
+private:
+    std::vector<double> _value;
+
+private:
+    ImageSize(std::vector<double> imageSize);
```
grp: Interesting, why a `std::vector<double>` versus `double width, double height`?

Libraries/acdriver/Sources/Options.cpp
```diff
@@ -73,7 +73,7 @@ parseArgument(std::vector<std::string> const &args, std::vector<std::string>::co
     } else if (arg == "--launch-image") {
         return libutil::Options::NextString(&_launchImage, args, it);
     } else if (arg == "--enable-on-demand-resources") {
-        return libutil::Options::MarkBool(&_enableOnDemandResources, arg);
+        return libutil::Options::NextBool(&_enableOnDemandResources, args, it, true);
```
grp: 👍

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -599,29 +626,35 @@ write() const
...
     header.bitmaps.bitmap_count = 1;
     header.bitmaps.payload_size = compressed_data_length;
 
+    if (header.pixel_format == car_rendition_value_pixel_format_jpeg ||
+        header.pixel_format == car_rendition_value_pixel_format_raw_data) {
+        extra_headers_size += sizeof(struct car_rendition_data_header_raw);
+    }
+
+    std::vector<uint8_t> output = std::vector<uint8_t>(rendition_header_size + extra_headers_size + compressed_data_length);
+    uint8_t *output_bytes = reinterpret_cast<uint8_t *>(output.data());;
```
grp: nit: two `;;`

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -564,15 +527,23 @@ write() const
...
-    if (nslices == 1) {
+    if (_slices.size() > 0) {
+        int index = 0;
+        info_slices->nslices = _slices.size();
+        for (auto const &slice : _slices) {
+            info_slices->slices[index].x = slice.x;
+            info_slices->slices[index].y = slice.y;
+            info_slices->slices[index].width = slice.width;
+            info_slices->slices[index].height = slice.height;
+            index += 1;
```
grp: nit: maybe use `for (size_t i = 0; i < _slices.size(); i++)` so both sides use the same indexing?

Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <xcassets/Asset/ImageSet.h>
+
+#include <acdriver/Result.h>
+
+#include <string>
+
+
+bool
+CompileContents(ext::optional<car::Writer> &writer,
```
grp: I like having separate files for each type of compilation. Maybe make a class for this so each type can be made with a consistent API? For example, `pbxbuild` has a set of `ToolResolver` classes for each tool in the build.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,490 @@
...
+            break;
+    }
+
+    std::vector<unsigned char> pixels;
+    size_t width = 0;
+    size_t height = 0;
+    size_t channels = 0;
+    car::Rendition::Data::Format format;
+
+    if (ends_with(filename, {std::string("png")})) {
```
grp: For this, try `FSUtil::GetFileExtension(filename) == "png"` (and same below for JPEG).

Libraries/acdriver/Sources/CompileAction.cpp
marcsalem: It's a quick re-write of the ContentsAction, it'll get cleaned up.

Libraries/acdriver/Headers/acdriver/Options.h
marcsalem: Good point, I'll switch to vector.

Libraries/xcassets/Headers/xcassets/Slot/ImageSize.h
marcsalem: I'll switch to width/height. The runtime warnings were bugging me so I just did a quick implementation.

Libraries/libcar/Sources/Rendition.cpp
marcsalem: Fixed, I cleaned up some of yours while I was at it :-p

Libraries/libcar/Sources/Rendition.cpp
marcsalem: sure

Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
marcsalem: Yeah, it I'm not super happy with this layout. I'll tidy it up with the next asset type that gets added.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Yeah, I should have checked FSUtil. Maybe it should also do the tolower transform.

marcsalem:
This also can't land until TravisCI whitelists libpng16. libpng12 is too old.

Libraries/xcassets/Headers/xcassets/Slot/ImageSize.h
marcsalem: LOL, yeah, guess where I hacked it from. It's cleaned up.

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -227,7 +188,9 @@ Load(
                         };
                         slices.push_back(slice);
                 }
-                rendition.slices() = std::move(slices);
+                if (slices.size() > 0) {
```
grp: What's this check do?

marcsalem: Ah. There should be a default slice that is the whole image. Right now that's a special case in the write method if the size is zero.

grp: Yes, but this doesn't seem to actually do anything. If the if statement wasn't there, rendition.slices() would still be empty. The only effect seems to avoid the copy?

marcsalem: The point is that I should be initializing the slices, and then I can take out the special case in the write function.

grp: Ah, ok, that makes sense.

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -250,6 +248,10 @@ Load(
     rendition.layout() = layout;
     rendition.resizeMode() = ResizeModeFromLayout(layout);
 
+    if (rendition.slices().size() == 0) {
+        rendition.slices().push_back({0, 0, value->width, value->height});
```
grp: What about doing this in the constructor? That way the `write()` method can write out exactly what's in the object, even if it's technically invalid, and it's still easy to get it right for clients.

grp:
Since most/all of the changes to `libcar` and related from this are merged now, it might make sense to make a new PR cherry-picking the commits for the `acdriver` changes. I think this one has grown pretty hard to review with lots of merging?

marcsalem:
Creating new pull request with content that has not been cherry-picked

PR #65, opened 2016-06-17
WIP Compile asset catalogs in acdriver.

grp:
This seems to be the mostly same as the last commit of #64, but at bit more developed in terms of output. Might be able to combine these and use the `CompileActionImageSet.cpp` from #64 along with the `CompileOutput` from this one.


marcsalem:
Nice! OK, I'll merge this into #64

Issue #3, opened 2015-11-05
Implement -usage command line option


grp:
xcbuild has a `-usage` option to explain the usage of the tool. Right now, this option is unimplemented.

To implement this, change `Driver.cpp` in `xcdriver` to call into `UsageAction::Run()`. Then, create a `UsageAction` class and implement the `Run()` method to print out usage info.

chessgeek:
I implemented this option recently, and I'm working on the -help option too. Are there any differences between this and the xcodebuild -usage option that are notable, or is it basically just a giant copy-paste?

grp:
I don't think so, although there are the extension options `-formatter`, `-executor`, and `-generate` which would be specific to xcbuild.

It is important to avoid copy/pasting directly from `xcodebuild`, since xcbuild is standalone, but a quick writeup of what the options do would be great!

chessgeek:
Thanks man! I'll get to work. I'm just starting to really dig into your codebase (found this project, like, a day ago), so I thought implementing the help and usage options would make me really have to understand what services it offers. Seems like some real cool stuff..!

PR #68, opened 2016-06-20
[xcdriver] Add `xcbuild -usage` action

modocache:
This addresses https://github.com/facebook/xcbuild/issues/3 by implementing a `-usage` action. The output mirrors `xcodebuild` exactly, although `xcbuild -usage` prints three additional options for each build action: `-formatter`, `-executor`, and `-generate`.

---

Let me know if this is what you had in mind. I wonder if we could automatically generate the lists of options, as opposed to hardcoding them... 🤔

### DIFF
--- a/Libraries/xcdriver/Sources/UsageAction.cpp
+++ b/Libraries/xcdriver/Sources/UsageAction.cpp
@@ -1,0 +1,93 @@
+/**
+ Copyright (c) 2016-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <xcdriver/UsageAction.h>
+
+using xcdriver::UsageAction;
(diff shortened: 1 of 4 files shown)


Libraries/xcdriver/Sources/UsageAction.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+}
+
+int UsageAction::
+Run(Filesystem const *filesystem, Options const &options)
+{
+    // TODO: Perhaps have xcdriver/Tools/xcbuild.cpp pass the name of the
+    //       executable (`argv[0]`) to `xcdriver::Driver::Run()`? That way,
+    //       the `UsageAction` could display the name of the program, instead
+    //       of hardcoding "xcbuild".
```
grp: You could use this for the equivalent of `argv[0]`:

    FSUtil::GetBaseName(SysUtil::GetExecutablePath());

Both `FSUtil` and `SysUtil` are in the `libutil` library.

grp:
(Updated with the mentioned fix and merged.)

modocache:
Sweet, thanks! 🙇

PR #69, opened 2016-06-21
Add ability to for xcbuild to read from a configuration file

Ktwu:
Xcode can be modified to include extra platform paths to find SDKs and extra toolchains using the 'defaults' utility.  xcbuild should have something similar.

Since xcbuild is cross-platform, I opted to use /var/db/xcbuild_defaults.plist as a configuration file.  In order to read from this file, plutil functionality has been split up and exposed as part of the plist library.

Accessing the configuration also all managed by the config library.  Right now it only has the ability to load the defaults plist file.


grp:
For the read/write functions, was there a reason you couldn't use `plist::Format::Any::Deserialize()` and `Filesystem::read()` or `plist::Format::Any::Serialize()` and `Filesystem::write()`? It doesn't special-case `"-"` for standard out, but that part seemed more specific to the implementation of plutil to me at least.

If `xcsdk::Manager::Open()` was passed arguments for the extra paths, that might make the manager more reusable (and easier to test) since it wouldn't read absolute file paths. Then the config loading part could be just a class to get the default config path and parse a config that you'd pass into to `xcsdk::Manager`. (At that point, it might even make sense to move it into `xcsdk`?)

Ktwu:
The read/write functions are just transplants from plutil.cpp -- I didn't bother trying to rewrite them.  I can upload another commit to do that if you'd like.

Passing in a config class to the manager makes sense -- I'll try that out.

Libraries/config/Sources/config.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <plist/Operations/Operations.h>
+
+#include <memory>
+#include <vector>
+
+std::unique_ptr<plist::Object>
+config::readDefaults(libutil::Filesystem const *filesystem) {
+   static std::string defaultConfigPath = "/var/db/xcbuild_defaults.plist";
+
+    std::pair<bool, std::vector<uint8_t>> result = plist::Read(filesystem, defaultConfigPath);
```
grp: This could probably just be:

    std::vector<uint8_t>> contents;
    if (!filesystem->read(&contents, defaultConfigPath)) {
         fprintf(stderr, "error: unable to read defaults\n");
         return nullptr;
    }

It seems like the same or less code as with `plist::Read()`, so I was wondering if that was needed at all. It doesn't look like you'd be using the functionality to interpret "-" specially, and `plist::Read()` doesn't do anything else.

Libraries/config/Sources/config.cpp
```diff
@@ -0,0 +1,40 @@
...
+        return nullptr;
+    }
+
+    auto format = plist::Format::Any::Identify(result.second);
+    if (format == nullptr) {
+        fprintf(stderr, "error: defaults file is not a plist\n");
+        return nullptr;
+    }
+
+    auto deserialize = plist::Format::Any::Deserialize(result.second, *format);
```
grp: You can skip the `plist::Format::Any::Identify()` call above here by just not passing the second parameter to `plist::Format::Any::Deserialize()`. It'll auto-identify the format if you don't specify it.

Libraries/config/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+add_library(config SHARED
+            Sources/Config.cpp
```
grp: Any reason this can't just go in `xcsdk`? One less library, I guess.

Libraries/config/Sources/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    return values;
+}
+
+Config::Config(libutil::Filesystem const *filesystem, std::string const &defaultConfigPath)
+{
+    if (char const *environConfigPath = getenv("CONFIG_PATH")) {
+        _path = std::string(environConfigPath);
```
grp: It's a little weird to have this override the parameter based on global state. If there was a `static` function on `Config` to get the default path that would then be passed into the load method, then the two operations could be separated out. That would let clients pick exactly what behavior they want.

Libraries/config/Headers/config/Config.h
```diff
@@ -0,0 +1,39 @@
...
+#define __config_config_h
+
+#include <libutil/Filesystem.h>
+#include <plist/Object.h>
+
+namespace config {
+
+class Config {
+public:
+    Config(libutil::Filesystem const *filesystem, std::string const &defaultConfigPath = "/var/db/xcbuild_defaults.plist");
```
grp: One thing to consider would be making this object just hold the *data* from the config, and making a static method like `Config::Load()` to actually load the configuration. So the constructor here would take `Config(std::vector<std::string> const &extraPlatformPaths, std::vector<std::string> const &extraToolchainPaths)` and just hold the data.

That way, you could easily create a config with the data you want, but it's still just one call to load the data. It also makes it easy for loading to fail — unlike a constructor, the `Load()` method could return `ext::optional<Config>` and `ext::nullopt` on failure.

Libraries/config/Sources/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+void Config::
+_load(libutil::Filesystem const *filesystem)
+{
+    _config = std::move(plist::Read(filesystem, _path));
+}
+
+std::vector<std::string> Config::
+ExtraPlatformPaths() const
+{
+    return _loadValue(_config.get(), "ExtraPlatformPaths");
```
grp: Is there a reason you need `plist::Extract` for this, versus just casting `_config` to a `plist::Dictionary *` and using `_config->value("ExtraPlatformPaths")`?

Libraries/pbxbuild/Sources/Build/Environment.cpp
```diff
@@ -52,7 +53,8 @@ Default(Filesystem const *filesystem)
      */
     specManager->registerDomains(filesystem, pbxspec::Manager::DefaultDomains(*developerRoot));
 
-    std::shared_ptr<xcsdk::SDK::Manager> sdkManager = xcsdk::SDK::Manager::Open(filesystem, *developerRoot);
+    auto config = std::unique_ptr<config::Config>(new config::Config(filesystem));
```
grp: Any reason to use `std::unique_ptr<Config>` here versus just leaving the `Config` on the stack?

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
```diff
@@ -88,7 +89,7 @@ class Manager {
     /*
      * Load from a developer root. Returns nullptr on error.
      */
-    static std::shared_ptr<Manager> Open(libutil::Filesystem const *filesystem, std::string const &path);
+    static std::shared_ptr<Manager> Open(config::Config const *config, libutil::Filesystem const *filesystem, std::string const &path);
```
grp: Any reason for `const *` versus a `const &`? Especially if `Config` becomes just a value, no need to pass by pointer AFAICT.

grp: nit: I'd put the `config` parameter last, after the `path`, since it's like additions to the path.

Libraries/config/CMakeLists.txt
Ktwu: Configurations seem like something any tool could benefit from, so I just made it a separate library.  And it seemed like distinct enough functionality from the rest of xcsdk.

Libraries/config/Sources/Config.cpp
Ktwu: My motivation for this approach is that the environment setting is an override no matter what library wants to use Config -- I feared a scenario where some future use of Config forgot to check the global environment first.  If you don't think that's reasonable, I'll change it.

Libraries/config/Headers/config/Config.h
Ktwu: That's a good idea.

Libraries/config/Sources/Config.cpp
Ktwu: Simplicity of use.  To me, it's cleaner to use ::Extract than casting and pulling the value from the dictionary.

Libraries/pbxbuild/Sources/Build/Environment.cpp
Ktwu: Nope.  I forgot C++ allows stack instances.

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
Ktwu: It's copy-paste from looking at the filesystem variable.  I'll change to a reference.

Libraries/config/Sources/Config.cpp
grp: I think if you made a function to return the default path rather than a parameter, that could check the environment. That way you can't forget but it's still separate from the loading path. Sort of like how `xcsdk::SDK::Manager::Load()` takes a path that `xcsdk::Environment::DeveloperRoot()` reads from the environment.

Ktwu: I'll do that, then.

Libraries/config/CMakeLists.txt
grp: Yeah, that's true. I guess my thought is that since it references platforms and toolchains in the configuration, that it's already sort of related.

I'm also not sure if configuration makes sense in general since it wouldn't be compatible with anything from Xcode generally. My thought was that coupling it closer might reduce the chances of using it when it's not appropriate if it requires more conscious effort.

Libraries/config/Sources/Config.cpp
grp: To me, they're for very different purposes: `Extract()` was intended to implement the very specific user-facing key-path syntax that `plutil` accepts.

I'm concerned that if it starts being used for programmatic purposes, other features then depend on the specific user-facing syntax of `plutil`. Any changes to that syntax for `plutil` might then break (for example) loading configs or anything else that uses it.

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
grp: It's a subtle distinction, and I'm not sure if it's common in C++ or just in this code. To me, a pointer "means" passing a resource while explicitly not copying it, where a const reference is just an optimization for passing by value.

Since a `Filesystem` is an expensive thing to copy (conceptually, copying the entire filesystem it represents would be ridiculous), a pointer makes sense. But especially if `Config` becomes a value type, there's no reason not to pas by value, and the const reference would just make that faster.

Ktwu: I've been told by an Oculus engineer that general styling is always to use references unless you absolutely need to use a pointer :)

Libraries/config/CMakeLists.txt
Ktwu: Ah, well, the code is already in a separate library; do you think it's worth moving it into xcsdk to cut down on the number of libraries?

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
grp: Yeah, there's definitely different standards. I personally like [Google's reasoning](https://google.github.io/styleguide/cppguide.html#Reference_Arguments) for avoiding non-const references: "References can be confusing, as they have value syntax but pointer semantics."

Libraries/config/CMakeLists.txt
grp: I would, but just because it's extra code. Maybe at some point it'll make sense to pull it out... If it's in `xcsdk`, could even call it something like `xcsdk::AdditionalPaths` to make it really separate from general config.

Libraries/config/Sources/Config.cpp
Ktwu: Ok.  It's not a big deal to change it, so I'll go ahead and do that.

Libraries/config/CMakeLists.txt
Ktwu: Do you mind if we leave this in for the moment, then?  I can follow up later, but my team is a bit blocked on this; I'd like to focus on necessary changes first and cleanup later.

Libraries/config/Headers/config/Config.h
Ktwu: Do you have any issue with just returning an empty Config option in the case of failure, or would you really like a ext::optional<Config>?

grp: Probably fine, although would be harder to test.

Libraries/config/Headers/config/Config.h
```diff
@@ -0,0 +1,39 @@
...
+
+public:
+    std::vector<std::string> ExtraPlatformPaths() const;
+    std::vector<std::string> ExtraToolchainPaths() const;
+
+public:
+    inline const std::string& path() const { return _path; }
```
grp: nit:

    std::string const &path() const
    { return _path; }

Libraries/config/Headers/config/Config.h
```diff
@@ -0,0 +1,39 @@
...
+
+private:
+    void _load(libutil::Filesystem const *filesystem);
+
+private:
```
grp: nit: members usually at the top

Libraries/config/CMakeLists.txt
grp: Probably fine, yeah.

grp:
I merged a slight variant in b29fda78f8d4b9cd26967bee8b4d0874a7469a19, with a few logic changes and unit tests. Note the paths, environment variables, and plist keys are slightly different from your code here: see the commit message for details.

Ktwu:
Thanks!

PR #70, opened 2016-06-21
Add `-help` action. Fixes #2.

### LINKED ISSUE #2: Implement -help command line option
xcbuild has a `-help` option to explain the usage of the tool. Right now, this option is unimplemented.

To implement this, change `Driver.cpp` in `xcdriver` to call into `HelpAction::Run()`. Then, create a `HelpAction` class and implement the `Run()` method to print out usage info.

modocache:
This addresses #2 by implementing a `-help` action. `-help` prints `-usage`, then lists explanations of each option. Options that are not yet implemented are clearly marked as such to avoid confusion.

### DIFF
--- a/Libraries/xcdriver/Headers/xcdriver/Usage.h
+++ b/Libraries/xcdriver/Headers/xcdriver/Usage.h
@@ -1,0 +1,21 @@
+/**
+ Copyright (c) 2016-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#ifndef __xcdriver_Usage_h
+#define __xcdriver_Usage_h
+
+#include <xcdriver/Base.h>
+
+namespace xcdriver {
+
+std::string usage();
+
+}
+
+#endif // !__xcdriver_Usage_h
--- a/Libraries/xcdriver/Sources/HelpAction.cpp
+++ b/Libraries/xcdriver/Sources/HelpAction.cpp
@@ -1,0 +1,222 @@
+/**
+ Copyright (c) 2016-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <xcdriver/HelpAction.h>
+#include <xcdriver/Usage.h>
+#include <libutil/FSUtil.h>
+#include <libutil/SysUtil.h>
+
+using xcdriver::HelpAction;
+using xcdriver::usage;
+using libutil::FSUtil;
+using libutil::SysUtil;
+
+HelpAction::
+HelpAction()
+{
+}
+
+HelpAction::
+~HelpAction()
+{
+}
+
+int HelpAction::
+Run()
+{
+    fprintf(stdout, "%s", usage().c_str());
+
+    fprintf(stdout, "Options:\n");
+    fprintf(
+        stdout,
+        "    -usage                                      "
+        "print brief usage instructions\n");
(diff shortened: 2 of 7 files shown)


Libraries/xcdriver/Sources/HelpAction.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+HelpAction::
+~HelpAction()
+{
+}
+
+int HelpAction::
+Run()
+{
+    UsageAction::Run();
```
grp: Interesting. I would probably structure this by pulling the usage code out into a separate class (that just returns the usage text as a string), then making both this action and the `HelpAction` print it. That way the `Action` classes have just the responsibility of performing the action from the command line, but not implementing the underlying functionality.

Libraries/xcdriver/Sources/HelpAction.cpp
```diff
@@ -0,0 +1,221 @@
...
+        "print complete usage instructions\n");
+    fprintf(
+        stdout,
+        "    -verbose                                    "
+        "print diagnostic messages and other detailed information when "
+        "performing an action\n");
+    fprintf(
+        stdout,
+        "    -license                                    "
+        "print xcbuild's license (not yet implemented)\n");
```
grp: nit: Might say just "print license" rather than hardcoding the name. (Should file a starter task for this one too, would be pretty easy to add!)

Libraries/xcdriver/Sources/HelpAction.cpp
```diff
@@ -0,0 +1,221 @@
...
+        "    -find-executable NAME                       "
+        "print the path to executable NAME in the provided SDK and toolchain\n");
+    fprintf(
+        stdout,
+        "    -find-library NAME                          "
+        "not yet implemented\n");
+    fprintf(
+        stdout,
+        "    -version                                    "
+        "print the version of xcbuild, or the version of any SDKs specified "
```
grp: nit: again might leave out "xcbuild"

modocache:
All good points, thanks! Will do tomorrow morning, but you can feel free to beat me to it :wink:

modocache:
@grp Updated, thanks for the tips! I've written about 20 lines of C++ in my life, so hopefully the `xcdriver::usage()` function is what you had in mind 😅 

I also updated the help text to no longer say `-license` was unimplemented, since you added it in 5d9f5583af2f06cab627b91f15ebfab856457cf3 -- thanks!

Libraries/xcdriver/Sources/Usage.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <libutil/SysUtil.h>
+#include <sstream>
+
+using libutil::FSUtil;
+using libutil::SysUtil;
+
+std::string xcdriver::usage() {
+	std::string executable = FSUtil::GetBaseName(SysUtil::GetExecutablePath());
+
+	std::ostringstream result;
```
grp: nit: these lines seem a little extra indented.

Also, in C++, `std::string` is mutable. Combined with how operator overloading interacts with references, that means that `+=` isn't any less efficient than using a string builder / string stream. That's unlike Java or Objective-C, where `StringBuilder` or `NSMutableString` would be more efficient than appending immutable strings repeatedly.

Libraries/xcdriver/Headers/xcdriver/Usage.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#ifndef __xcdriver_Usage_h
+#define __xcdriver_Usage_h
+
+#include <xcdriver/Base.h>
+
+namespace xcdriver {
+
+std::string usage();
```
grp: This works, but in here the more common convention is to put this in a class as a static method:

    class Usage {
    private:
        Usage();
        ~Usage();
    
    public:
        static std::string Text();
    };

The `private:` block is to make the class purely static — since the constructor is private, there's no way to create an instance of it. That's all good, though, since the only method on it is `Usage::Text()` and it's static.

PR #71, opened 2016-06-22
[xcbuild] Fixes to build on Linux

Ktwu:
For whatever reason, xcbuild couldn't find unique_ptr, strcmpcase, and other basic std:: functionality unless I included the headers.


grp:
Sadly it looks like Travis is very backed up right now: https://travis-ci.org/facebook/xcbuild/builds

PR #72, opened 2016-06-23
Add imageset support in Compile Action

marcsalem:
Implements actool Compile action with:
  * imageset supported
  * PNG converted to zlib compressed PremultipliedBGRA8 or PremultipliedGA8
  * JPEG/raw support
  * size, scale, idiom supported
  * resizing partially supported
  * adds requirement for libpng

### DIFF
--- a/.travis.linux.trusty.sh
+++ b/.travis.linux.trusty.sh
@@ -1,0 +1,21 @@
+#! /bin/bash
+
+URL=ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.23.tar.gz
+FILE=libpng-1.6.23.tar.gz
+MD5SUM=a49e4cc48d968c79def53d082809c9f2
+
+echo Downloading $URL
+curl $URL -o $FILE
+echo "$MD5SUM  $FILE" | md5sum -c
+if [ "$?" != "0" ]; then
+    exit 1
+fi
+
+echo Extracting
+tar -zxvvf $FILE
+cd ${FILE%.tar.gz}
+
+echo Running configure/make/make install
+./configure --prefix=/usr/local
+make
+make install
--- a/Libraries/acdriver/CMakeLists.txt
+++ b/Libraries/acdriver/CMakeLists.txt
@@ -14,12 +14,17 @@ add_library(acdriver SHARED
             Sources/Result.cpp
             Sources/VersionAction.cpp
             Sources/CompileAction.cpp
+            Sources/CompileActionImageSet.cpp
             Sources/CompileOutput.cpp
             Sources/ContentsAction.cpp
             )
 
+find_package(PNG REQUIRED)
+
 target_link_libraries(acdriver PUBLIC xcassets util plist dependency ext bom car ${PNG_LIBRARIES})
 target_include_directories(acdriver PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Headers")
+target_include_directories(acdriver PRIVATE ${PNG_INCLUDE_DIRS})
+set_target_properties(acdriver PROPERTIES COMPILE_DEFINITIONS "${PNG_DEFINITIONS}")
 
 install(TARGETS acdriver DESTINATION usr/lib)
 
--- a/Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
+++ b/Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
@@ -20,9 +20,9 @@
 
 bool
 CompileAsset(
-    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    xcassets::Asset::ImageSet::Image const &image,
+    std::shared_ptr<xcassets::Asset::Asset> const &parent,
+    libutil::Filesystem *filesystem,
     acdriver::Options const &options,
     acdriver::CompileOutput *compileOutput,
-    acdriver::Result *result,
-    std::shared_ptr<xcassets::Asset::Asset> const &parent,
-    xcassets::Asset::ImageSet::Image const &image);
+    acdriver::Result *result);
--- a/Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
+++ b/Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
@@ -1,0 +1,28 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <acdriver/CompileOutput.h>
+#include <acdriver/Result.h>
+
+#include <ext/optional>
+#include <car/Writer.h>
+
+#include <xcassets/Asset/Catalog.h>
+#include <xcassets/Asset/ImageSet.h>
+
+#include <string>
+
+bool
+CompileAsset(
+    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    acdriver::Options const &options,
+    acdriver::CompileOutput *compileOutput,
+    acdriver::Result *result,
+    std::shared_ptr<xcassets::Asset::Asset> const &parent,
+    xcassets::Asset::ImageSet::Image const &image);
--- a/Libraries/acdriver/Sources/CompileAction.cpp
+++ b/Libraries/acdriver/Sources/CompileAction.cpp
@@ -8,14 +8,49 @@
  */
 
 #include <acdriver/CompileAction.h>
+#include <acdriver/CompileActionImageSet.h>
+#include <acdriver/CompileOutput.h>
 #include <acdriver/Options.h>
 #include <acdriver/Output.h>
 #include <acdriver/Result.h>
+#include <xcassets/Asset/AppIconSet.h>
+#include <xcassets/Asset/BrandAssets.h>
+#include <xcassets/Asset/Catalog.h>
+#include <xcassets/Asset/ComplicationSet.h>
+#include <xcassets/Asset/DataSet.h>
+#include <xcassets/Asset/GCDashboardImage.h>
+#include <xcassets/Asset/GCLeaderboard.h>
+#include <xcassets/Asset/GCLeaderboardSet.h>
+#include <xcassets/Asset/Group.h>
+#include <xcassets/Asset/IconSet.h>
+#include <xcassets/Asset/ImageSet.h>
+#include <xcassets/Asset/ImageStack.h>
+#include <xcassets/Asset/ImageStackLayer.h>
+#include <xcassets/Asset/LaunchImage.h>
+#include <xcassets/Asset/SpriteAtlas.h>
+#include <xcassets/Slot/Idiom.h>
+#include <libutil/Filesystem.h>
+#include <libutil/FSUtil.h>
+#include <bom/bom.h>
+#include <car/Reader.h>
+#include <car/Writer.h>
+#include <bom/bom_format.h>
+#include <dependency/BinaryDependencyInfo.h>
+#include <plist/Array.h>
+#include <plist/Boolean.h>
+#include <plist/Dictionary.h>
+#include <plist/String.h>
 
 using acdriver::CompileAction;
+using acdriver::CompileOutput;
 using acdriver::Options;
 using acdriver::Output;
 using acdriver::Result;
+using libutil::Filesystem;
+
+using libutil::Filesystem;
+using libutil::FSUtil;
+
 
 CompileAction::
 CompileAction()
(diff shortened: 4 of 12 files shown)


Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+{
+    char header[8];
+    png_struct *png_struct_ptr;
+    png_info *info_struct_ptr;
+
+    int number_of_passes;
+    png_uint_32 row_bytes;
+    png_byte **row_pointers = NULL;
+
+    FILE *fp = fopen(filename, "rb");
```
grp: Can you use the `Filesystem` abstraction for this, so it can be unit tested? You'd need to load the PNG from memory.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+    }
+    if (channels_out) {
+        *channels_out = channels;
+    }
+    return true;
+}
+
+static uint16_t
+GenerateIdentifier(void) {
+    static uint16_t last = 0;
```
grp: It's a little weird that multiple catalogs compiled in the same process would behave differently than in different processes. Maybe attach state to the compilation?

Libraries/acdriver/Sources/Options.cpp
```diff
@@ -19,6 +19,7 @@ Options() :
     _errors(false),
     _notices(false),
     _compressPNGs(false),
+    _targetDevice({}),
```
grp: This isn't necessary — `std::vector` has a default constructor that makes an empty one.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+    std::string filename = parent->path() + std::string("/") + *image.fileName();
+
+    /* An image without an idiom is considered unassigned */
+    if (!image.idiom()) {
+        return false;
+    }
+
+    /* Add namespace to name */
+    if (ns.size() > 0) {
+        name = ns + std::string("/") + name;
```
grp: Instead of passing around and building up the namespaced asset name, you can use `asset->name().string()`. It's already built up while loading into an `xcassets::FullyQualifiedName`.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+    /* Add namespace to name */
+    if (ns.size() > 0) {
+        name = ns + std::string("/") + name;        
+    }
+
+    /* scale defaults to 0 / all */
+    double scale = 0;
+    if (image.scale()) {
+        scale = image.scale()->value();
+    }
```
grp: Instead of the conditional, `ext::optional` has a method to do it in one condition: `image.scale().value_or(0)`.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+        case xcassets::Slot::Idiom::TV:
+            idiom = car_attribute_identifier_idiom_value_tv;
+            break;
+        case xcassets::Slot::Idiom::Watch:
+            idiom = car_attribute_identifier_idiom_value_watch;
+            break;
+        case xcassets::Slot::Idiom::Car:
+            idiom = car_attribute_identifier_idiom_value_car;
+            break;
+    }
```
grp: This could probably be a separate function.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+    last = last + 1;
+    return last;
+}
+
+bool
+CompileAsset(
+    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    std::string ns,
+    Options const &options,
+    CompileOutput *compileOutput,
```
grp: One of the output parameters `acdriver::CompileOutput::format()` is a format: in some cases, the asset catalog shouldn't be built, and it should instead just copy the images into the output.

I think it wouldn't be too hard to support that, although it'd require transforming the asset options into filename suffixes. If that's a separate PR, maybe make it warn for now?

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -0,0 +1,498 @@
...
+    std::string name = FSUtil::GetBaseNameWithoutExtension(parent->path());
+
+    /* Skip any entry that is not attached to a file */
+    if (!image.fileName()) {
+        return true;
+    }
+
+    std::string filename = parent->path() + std::string("/") + *image.fileName();
+
+    /* An image without an idiom is considered unassigned */
```
grp: Might want to explicitly check `image.unassigned()` too?

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Yeah, ordering will change the ID, but does that really matter? I could make it a hash of the name including the namespace, but it's still have to have a map to avoid collisions.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: sure

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Ah, I'll add a warning for now.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Ok

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: For me, parent->name().string() only returns the name without the namespace, and groups() is empty. I'll debug it.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
grp: I like the incremental IDs. Maybe just pass down the counter from the `CompileAction` (or put it on `CompileOutput` which is already passed down) so it's deterministic?

.travis.linux.trusty.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/bash
+
+URL=ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.23.tar.gz
+FILE=libpng-1.6.23.tar.gz
```
grp: Would it make more sense to include libpng as a submodule? Then it's easily just compiled into the build, rather than done separately. Or, even do it this way: in CMake, run `FindLibPNG`, and if it doesn't find it, then include the submodule version.

Libraries/acdriver/CMakeLists.txt
```diff
@@ -14,11 +14,17 @@ add_library(acdriver SHARED
             Sources/Result.cpp
             Sources/VersionAction.cpp
             Sources/CompileAction.cpp
+            Sources/CompileActionImageSet.cpp
+            Sources/CompileOutput.cpp
```
grp: I think it'd probably be good to add this in a separate commit (maybe even PR) with tests. Should be straightforward to test, I think.

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -27,10 +62,312 @@ CompileAction::
...
+
+static std::string
+AssetReference(std::shared_ptr<xcassets::Asset::Asset> const &asset)
+{
+    // TODO: include [] for each key
+    return asset->path();
+}
+
+static void
+CompileAppIconSet(std::shared_ptr<xcassets::Asset::AppIconSet> const &appIconSet, CompileOutput *compileOutput, Result *result)
```
grp: This is incomplete enough it should probably be a separate PR after this one goes in, I think.

Libraries/libutil/Tests/test_FSUtil.cpp
```diff
@@ -83,3 +83,21 @@ TEST(FSUtil, ResolveRelativePath)
     EXPECT_EQ("c/a/b", FSUtil::ResolveRelativePath("a/b", "c"));
     EXPECT_EQ("/a/b", FSUtil::ResolveRelativePath("/a/b", "/c"));
 }
+
+TEST(FSUtil, IsFileExtension)
```
grp: Maybe I'm confused, but where was this added? Is the implementation in a later commit?

grp: Oh, ha. Looks like it's been there all along.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -259,10 +259,7 @@ CompileAsset(
     std::string name = parent->name().string();
 
     /* scale defaults to 0 / all */
-    double scale = 0;
-    if (image.scale()) {
-        scale = image.scale()->value();
-    }
+    double scale = image.scale()->value();
```
grp: Oh, my bad, this won't work. If `image.scale()` doesn't exist, it'll be `ext::nullopt` and crash on dereference. I thought that was of type `ext::optional<double>`, not `ext::optional<xcassets::Slot::Scale>`. Sorry!

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
```diff
@@ -29,6 +29,7 @@ class Asset {
 private:
     FullyQualifiedName         _name;
     std::string                _path;
+    libutil::Filesystem const *_filesystem;
```
grp: Hm, this is a little weird. First, though, I agree that it makes sense that the same filesystem should necessarily be the one used to read it later.

However... this does create a sort of strange situation in two ways. First, a data object is holding onto a resource, which conceptually makes the object more complex than just being immutable data. Second, the lifetime of the filesystem is unspecified in the API, and it's up to clients of this class to keep the filesystem alive at least as long as the asset is, which is definitely non-obvious. Rust's lifetimes could enforce that at compile time, but lacking those in C++ I think it's better to avoid the complexity and pass around the filesystem separately where it's needed.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Either way it will still depend on the filesystem, and the order directory contents are read.

.travis.linux.trusty.sh
marcsalem: I thought about that, but really most people can use the system libpng. The image used by TravisCI is just really old at this point.

Libraries/acdriver/CMakeLists.txt
marcsalem: I don't see why this needs to be split up. Are there any existing tests in xcbuild for CLIs?

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
marcsalem: I'm just using this in the Catalog object.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
grp: True, although that is fixed with `libutil::MemoryFilesystem` as you'd use in tests.

.travis.linux.trusty.sh
grp: A submodule is pretty similar to what you're doing here, it's just avoiding a different build system. I think that's nicer especially for OS X users since they're unlikely to have libpng installed into the system. But if you check `FindLibPNG` first, then it can still use the system one if available...

Libraries/acdriver/CMakeLists.txt
grp: Yeah, definitely. The tests for `acdriver` tests `Result`, `Options`, and `Output` already.

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
grp: Hm, where in the catalog object? If it's synchronous with creation, you should be able to just pass it down as a parameter. If it's later on (e.g. a method) then it still has the same issues about requiring a specific lifetime for the filesystem.

.travis.linux.trusty.sh
marcsalem: The libpng api is pretty stable, this pull request just happens to use a newer feature to get the bitmap in one of the formats expected by libcar. I think it's fine to use the system provided version as long as it is 1.6+ which is common enough by now.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
marcsalem: Well, as long as MemoryFilesystem is consistent with ordering, it seems fine to me. A hashing scheme based on the name sounds fine also, but again, with only 16 bits, resolving collisions in one pass will probably also have different results based on the ordering of directory contents.

.travis.linux.trusty.sh
Ktwu: ++submodule approach.

Libraries/acdriver/Headers/acdriver/CompileActionImageSet.h
```diff
@@ -0,0 +1,29 @@
...
+
+bool
+CompileAsset(
+    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    std::string ns,
+    acdriver::Options const &options,
+    acdriver::CompileOutput *compileOutput,
+    acdriver::Result *result,
+    std::shared_ptr<xcassets::Asset::Asset> const &parent,
+	xcassets::Asset::ImageSet::Image const &image);
```
Ktwu: nit:  spacing

Libraries/acdriver/Headers/acdriver/CompileOutput.h
```diff
@@ -0,0 +1,117 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <plist/Dictionary.h>
```
Ktwu: nit: sort these

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
marcsalem: the filesystem object is passed to run(), run() creates the catalog, and passes that to CompileAction() and does not return it, and would not use it past the write() call at the end of run(). I don't see why I can't expect the filesystem to have the same lifetime as the Catalog object.

.travis.linux.trusty.sh
marcsalem: OK, which git repo?

Libraries/acdriver/CMakeLists.txt
```diff
@@ -14,11 +14,17 @@ add_library(acdriver SHARED
...
             Sources/ContentsAction.cpp
             )
 
-target_link_libraries(acdriver PUBLIC xcassets util plist ext)
+find_package(PNG REQUIRED)
+
+target_link_libraries(acdriver PUBLIC xcassets util plist dependency ext bom car ${PNG_LIBRARIES})
 target_include_directories(acdriver PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Headers")
+target_include_directories(acdriver PRIVATE "${PNG_INCLUDE_DIRS}")
```
strager: `PNG_INCLUDE_DIRS` should not be quoted.

strager: Nit: you're missing `PNG_DEFINITIONS`.

Libraries/acdriver/CMakeLists.txt
marcsalem: I don't see tests for the catalog object, while there is a ContentsAction that loads it and prints it.

I do have higher level tests that go from json -> car file -> dump_car to check that everything stays consistent. I'll follow up with something in a later pull request that has this coverage with a faked out filesystem.

grp: Yeah, I'm not saying tests for the compile action are needed right now, that's pretty complex since it would need mock assets as an input. But I think the `CompileOutput` should get tests since it's relatively much simpler.

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
grp: In that case, you can, but I can imagine other uses of the same objects that don't. For example, imagine if you made an asset catalog editor UI, you'd probably want to hold onto `xcassets::Asset::Asset` objects for a long time but the filesystem might be transient from the open dialog.

Of course it'd be possible to restructure that to work with this, but it requires understanding that implementation detail of the `xcassets::Asset::Asset` in order to use it safely. If the filesystem is passed in separately to the functions that need to use it, then you can't get it wrong.

Libraries/acdriver/CMakeLists.txt
```diff
@@ -23,7 +23,8 @@ find_package(PNG REQUIRED)
 
 target_link_libraries(acdriver PUBLIC xcassets util plist dependency ext bom car ${PNG_LIBRARIES})
 target_include_directories(acdriver PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Headers")
-target_include_directories(acdriver PRIVATE "${PNG_INCLUDE_DIRS}")
+target_include_directories(acdriver PRIVATE ${PNG_INCLUDE_DIRS})
+add_definitions(${PNG_DEFINITIONS})
```
grp: nit: the more fine-grained way to do this is `set_target_properties()` with `COMPILE_DEFINITIONS`

.travis.linux.trusty.sh
grp: http://libpng.sourceforge.net links to https://github.com/glennrp/libpng, which has a `libpng16` branch. the current `HEAD` of that branch is probably a good target for the submodule?

grp: You'll probably want to also set `PNG_STATIC` before adding the subdirectory (and disable `PNG_SHARED` and `PNG_TESTS`) to statically link it.

grp:
I pulled out the compile output part and added a test in https://github.com/facebook/xcbuild/pull/79. Still needs this PR on top in order to make use of it, though.

Libraries/xcassets/Headers/xcassets/Asset/Asset.h
marcsalem: What do you expect to change in the filesystem during the execution of the application?

I think it would be much better for Asset objects to have an API to return a byte vector or stream with the file contents. The only reason to know the path/filename is for the filename field in the CAR file which is useful for debugging.

I think it's awkward to have to pass around two arguments down to every function that works on assets, which is really each one that is not a stub, or not just a print contents action.

Libraries/acdriver/CMakeLists.txt
marcsalem: Sure, but there is no PNG_DEFINITIONS when I run cmake.

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -27,10 +62,312 @@ CompileAction::
 {
 }
 
+static bool
+CompileAsset(
+    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    std::string ns,
```
Ktwu: nit:  ns?  What's that?  Can you give a more descriptive name?

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -27,10 +62,303 @@ CompileAction::
...
+    CompileOutput *compileOutput,
+    Result *result,
+    std::shared_ptr<xcassets::Asset::Asset> const &parent,
+    std::vector<T> const &assets)
+{
+    bool success = true;
+
+    for (auto const &asset : assets) {
+        if (!CompileAsset(catalog, options, compileOutput, result, parent, asset)) {
+            success = false;
```
Ktwu: Are you purposefully trying to compile each asset even if one of them fails?

You could also do:

`success &= CompileAsset(...)`

marcsalem: This is from merging in Grants pull request:
https://github.com/facebook/xcbuild/pull/65/files

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -27,10 +62,303 @@ CompileAction::
 {
 }
 
+static bool
+CompileAsset(
+    std::shared_ptr<xcassets::Asset::Catalog> const &catalog,
+    Options const &options,
+    CompileOutput *compileOutput,
```
Ktwu: Are `result` and `compileOutput` considered as out parameters here?  If so, maybe they ought to move behind `parent` and `asset`.

.travis.linux.trusty.sh
Ktwu: strager, myself, and Marc talked about this offline; given that this is only useful for Travis, we're OK with downloading libpng as it is right now.

Libraries/acdriver/Sources/CompileActionImageSet.cpp
```diff
@@ -296,7 +297,7 @@ CompileAsset(
 
     if (FSUtil::IsFileExtension(filename, "png", true)) {
         std::vector<uint8_t> contents;
-        if (!catalog->filesystem() || !catalog->filesystem()->read(&contents, filename)) {
+        if (!filesystem || !filesystem->read(&contents, filename)) {
```
grp: Even though pointers can be `nullptr`, I wouldn't bother checking for it — there's clearly no meaning to this function without a value provided, and if there was one, it'd be an `ext::optional<Filesystem const *>` anyway.

.travis.linux.trusty.sh
grp: My concern is for OS X — I don't see libpng on my system. Unless I'm looking in the wrong places, where would a system one come from there? Required package manager installations (e.g. Homebrew, MacPorts, etc) means a dependency on a big third-party system that might not always be available.

marcsalem: libpng is a pretty standard stable library. There are lots of ways people can install it.

A more appropriate solution is to add an alternate code path for macOS which uses a built in library for reading PNG files, and maybe PDF too!

I also like Brian's idea to make a Homebrew formula, similarly we can make Ubuntu packages, and maybe even a snap package. None of these things require adding a submodule.

grp: Yeah; I'm only proposing the submodule as a fallback for systems that don't have libpng available otherwise. It's very similar to what you have here but in a way that works off of Travis too.

Ktwu:
Marc, Ted, and I talked about this offline; both Ted and I agreed that this code looked good-as is, and any remaining issues -- how to handle libpng, id generation, tests, and stubbed functionality -- shouldn't hold this patch back (especially since it's in active development).

lgtm

.travis.linux.trusty.sh
narissiam: Ran into an issue where actool fails to load libpng16 on CentOS 7, which has libpng15 as the default. I think having libpng as a submodule would be worth revisiting.

PR #73, opened 2016-06-24
[xcbuild] Restore install behavior for libs and exes

Ktwu:
cc strager who can best defend the reason for this.


strager:
Discussion on the blamed revision: https://github.com/facebook/xcbuild/commit/c25336b50b7735c50a692f2453b6dded435ceb14#commitcomment-18010614

PR #75, opened 2016-06-25
[xcbuild] Split Libraries/Specifications into separate install components

Ktwu:
Using components allows builds to install only the stuff it needs.

### DIFF
--- a/Libraries/CMakeLists.txt
+++ b/Libraries/CMakeLists.txt
@@ -7,6 +7,9 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
+set(OLD_DEFAULT_COMPONENT ${CMAKE_INSTALL_DEFAULT_COMPONENT_NAME})
+set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME libraries)
+
 add_subdirectory(acdriver)
 add_subdirectory(builtin)
 add_subdirectory(dependency)
@@ -27,4 +30,6 @@ add_subdirectory(xcformatter)
 add_subdirectory(xcscheme)
 add_subdirectory(xcsdk)
 add_subdirectory(xcworkspace)
+
+set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME ${OLD_DEFAULT_COMPONENT})
 
--- a/Specifications/CMakeLists.txt
+++ b/Specifications/CMakeLists.txt
@@ -7,6 +7,9 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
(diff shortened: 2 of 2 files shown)


Libraries/acdriver/CMakeLists.txt
```diff
@@ -19,11 +19,11 @@ add_library(acdriver SHARED
 
 target_link_libraries(acdriver PUBLIC xcassets util plist ext)
 target_include_directories(acdriver PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Headers")
-install(TARGETS acdriver DESTINATION usr/lib)
```
strager: Instead, push/set/pop `CMAKE_INSTALL_DEFAULT_COMPONENT_NAME` in the `Libraries/CMakeLists.txt` (or whatever calls `add_subdirectory`). This reduces the diff and is less nasty (though is action-at-a-distance).

Ktwu: Sure, that seems reasonable.

Ktwu:
Any feedback?  Is this a good idea?  Bad idea?

strager:
From a CMake language perspective, I think this diff looks good. I'll let a maintainer approve.

Why are the specifications optionally installed?

Ktwu:
> Why are the specifications optionally installed?

What do you mean optionally installed?  If no component is specified, all are installed by default.

strager:
A CMake user can choose to not install specifications. Why would a CMake user do that?

Ktwu:
I know little about these specifications, but they appear related to Xcode, right?  I don't have Xcode on Linux, my target environment, so I don't care about specifications.

grp:
Specifications are used by xcbuild (and Xcode) in order to resolve build settings into command line flags. If you want to do builds on Linux using xcbuild, you'll need a set of specifications — either the ones from Xcode or the (still incomplete) set from xcbuild.

Ktwu:
Ok, that's good to know.  For the moment, all I actually care about are some of the tools inside of xcbuild (plutil on Linux, for example) -- I don't need specifications for each of the individual tools, do I?

grp:
Nope, just for building.

Ktwu:
Cool.  So, to answer strager's question, I care about this because I actually only want xcbuild's tooling.  I'm not using xcbuild itself because I'm trying to integrate the use of these tools with build structures set up with BUCK.

strager:
Mmk, lgtm.

Libraries/acdriver/CMakeLists.txt
grp: I kinda like the way @Ktwu did it first, but not worth extra changes now.

PR #76, opened 2016-06-25
[xcsdk] Treat TOOLCHAINS as a list of possible toolchains, not just one toolchain

Ktwu:
TOOLCHAINS should be a space-delimited string of toolchain names.  xcrun
now correctly parses this string to build a final toolchains list.

### DIFF
--- a/Libraries/xcsdk/Tools/xcrun.cpp
+++ b/Libraries/xcsdk/Tools/xcrun.cpp
@@ -16,6 +16,7 @@
 #include <libutil/Options.h>
 #include <libutil/Subprocess.h>
 #include <libutil/SysUtil.h>
+#include <pbxsetting/Type.h>
 
 using libutil::DefaultFilesystem;
 using libutil::Filesystem;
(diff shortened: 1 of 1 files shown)


Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -324,15 +327,27 @@ main(int argc, char **argv)
...
         /* If the custom toolchain exists, use it instead. */
-        if (auto TC = manager->findToolchain(toolchain)) {
-            toolchains = { TC };
-        } else {
-            fprintf(stderr, "error: unable to find toolchain '%s'\n", toolchain.c_str());
+        std::vector<std::string> toolchainTokens;
+        std::stringstream sstream(toolchainsInput);
+        copy(std::istream_iterator<std::string>(sstream),
+             std::istream_iterator<std::string>(),
+             std::back_inserter(toolchainTokens));
```
grp: You could use `pbxsetting::Type::ParseList()` here, it's a little shorter and IMO more clear about what it's doing.

Ktwu: Nice.  I didn't know about this function; I'll definitely use it.

PR #78, opened 2016-06-27
Support preprocessing Info.plist before the Info.plist utility.

### LINKED ISSUE #77: Support INFOPLIST_PREPROCESS and related build settings
I noticed that some of the property list files being built by xcbuild are not formatted correctly. I found a few example property lists that show generic examples that most developers will try to use in preprocessing of these files. In general this preprocessor works as a C Preprocessor with a few quirks added. Some notes on this can be found on Apple [Technical Note TN2175](https://developer.apple.com/library/prerelease/content/technotes/tn2175/_index.html). 

In general I would assume that the plist utility will support the following operations...
- Pre-processor Conditionals:  #if, #ifdef, #ifndef, #else, #elif and #endif.
- C-style Relational Operators ( see: [Wikipedia: Operators in C and C++](https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B)  )
- C-style Logical Operators ( see: [Wikipedia: Operators in C and C++](https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B)  )
- C-Style comments ``` /* Comment */```
- C++-Style comments ``` // Comment ``` ( when -traditional is not set )

Example 1: IOAudioFamily version 204.3
This is the one of the simplest examples of plist editing. All it does is define a few variables in an xcode configuration file and output it in the kernel extensions PLIST file..

Example 2: IOHIDFamily version 701.20.10
A slightly more complex example. This configures multiple property lists.

Example 3: IOStorageFamily version 191
This includes an example of preprocessor events to remove sections of the property list on a Embedded platform where TARGET_OS_EMBEDDED is defined in the TargetConditionals header.

Example 4: Some more Pre-processor examples.
For the most part if it's in the TargetConditionals header for OSX, it'll appear in the Property list utility.
 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>PreProcess_KEYS</key>
	<dict>
		<key>IsEmbedded</key>
#if !TARGET_OS_EMBEDDED
		<true/>
#else 
		<false/>
#endif

		<key>IsMac</key>
#if TARGET_OS_MAC
		<true/>
#else 
		<false/>
#endif
		<key>IsPowerPC</key>
#if TARGET_CPU_PPC
		<true/>
#else 
		<false/>
#endif
	</dict>
</dict>
```

grp:
The `INFOPLIST_PREPROCESS` build setting requests preprocessing of
the Info.plist file before standard processing. Fixes #77.

### DIFF
--- a/Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
+++ b/Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
@@ -306,7 +306,24 @@ resolve(Phase::Environment const &phaseEnvironment, Phase::Context *phaseContext
         std::string infoPlistFile = environment.resolve("INFOPLIST_FILE");
         if (!infoPlistFile.empty()) {
             if (pbxsetting::Type::ParseBoolean(environment.resolve("INFOPLIST_PREPROCESS"))) {
-                // TODO(grp): Preprocess Info.plist using configuration from other build settings.
+                if (Tool::ToolResolver const *toolResolver = phaseContext->toolResolver(phaseEnvironment, "com.apple.compilers.cpp")) {
+                    std::string infoPlistIntermediate = environment.resolve("TEMP_DIR") + "/" + "Preprocessed-Info.plist";
+
+                    pbxsetting::Level level = pbxsetting::Level({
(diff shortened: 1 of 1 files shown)


grp:
Any comments or should I just merge? cc @marcsalem who suggested more code review

PR #81, opened 2016-07-06
[RFC] Install support files in Developer/ and user programs in bin/

strager:
See individual commit messages for details.

Specific questions for the audience:

* Does supporting xcrun without an SDK make sense? (See "HACK/WIP Support xcrun without an SDK." commit.)
* I didn't realize xcrun won't necessarily use xcbuild's Developer directory. This makes the behaviour of the tools in ${CMAKE_INSTALL_PREFIX}/bin confusing to me. Is that a known problem or is it expected?
* Should we just install the user programs directly (duplicating them) instead of making aliases?
* Should I put the user programs in their own CMake install component? Currently they are in the "libraries" component.
* Will there ever be plans to support xcbuild as a set of APIs instead of a set of executables?
* What's a better name for "normal" installed files? I use "USER".

### DIFF
--- a/CMake/Install.cmake
+++ b/CMake/Install.cmake
@@ -7,9 +7,11 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
+set(XCBUILD_INSTALL_DEVELOPER_DIR Developer CACHE STRING "Install path for xcbuild support files")
+
 # XCBUILD_INSTALL_TARGET(foo bar qix) installs the CMake targets 'foo', 'bar',
 # and 'qix' into the xcbuild Developer dir (specified by
-# CMAKE_INSTALL_PREFIX).
+# XCBUILD_INSTALL_DEVELOPER_DIR).
 function (XCBUILD_INSTALL_TARGET)
   cmake_parse_arguments("" "" "" "" ${ARGN})
   set(TARGETS ${_UNPARSED_ARGUMENTS})
@@ -17,6 +19,6 @@ function (XCBUILD_INSTALL_TARGET)
   # libraries (including Windows import libraries).
   # TODO(strager): Handle FRAMEWORK and BUNDLE targets.
   install(TARGETS ${TARGETS}
-          RUNTIME DESTINATION usr/bin
-          LIBRARY DESTINATION usr/lib)
+          RUNTIME DESTINATION ${XCBUILD_INSTALL_DEVELOPER_DIR}/usr/bin
+          LIBRARY DESTINATION ${XCBUILD_INSTALL_DEVELOPER_DIR}/usr/lib)
 endfunction ()
--- a/CMake/Install.cmake
+++ b/CMake/Install.cmake
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2015-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+# XCBUILD_INSTALL_TARGET(foo bar qix) installs the CMake targets 'foo', 'bar',
+# and 'qix' into the xcbuild Developer dir (specified by
+# CMAKE_INSTALL_PREFIX).
+function (XCBUILD_INSTALL_TARGET)
+  cmake_parse_arguments("" "" "" "" ${ARGN})
+  set(TARGETS ${_UNPARSED_ARGUMENTS})
+  # ARCHIVE is deliberately omitted; intentionally do not install static
+  # libraries (including Windows import libraries).
+  # TODO(strager): Handle FRAMEWORK and BUNDLE targets.
+  install(TARGETS ${TARGETS}
+          RUNTIME DESTINATION usr/bin
+          LIBRARY DESTINATION usr/lib)
+endfunction ()
--- a/CMake/Install.cmake
+++ b/CMake/Install.cmake
@@ -12,8 +12,24 @@ set(XCBUILD_INSTALL_DEVELOPER_DIR Developer CACHE STRING "Install path for xcbui
 # XCBUILD_INSTALL_TARGET(foo bar qix) installs the CMake targets 'foo', 'bar',
 # and 'qix' into the xcbuild Developer dir (specified by
 # XCBUILD_INSTALL_DEVELOPER_DIR).
+#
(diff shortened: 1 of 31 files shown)


Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -236,15 +241,25 @@ Version()
...
      * Parse out the options, or print help & exit.
      */
     Options options;
-    std::pair<bool, std::string> result = libutil::Options::Parse<Options>(&options, args);
-    if (!result.first) {
-        return Help(result.second);
+
+    {
+        std::vector<std::string> args = std::vector<std::string>(argv + 1, argv + argc);
+        std::string xcrunBaseName = libutil::FSUtil::GetBaseName(argv[0]);
```
grp: There's a more complete version of `argv[0]` as `SysUtil::GetExecutablePath()`.

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -236,15 +241,25 @@ Version()
...
+        /* HACK(strager): There must be a better way. */
+        bool isXcrun = xcrunBaseName == "xcrun";
+        if (isXcrun) {
+            std::pair<bool, std::string> result = libutil::Options::Parse<Options>(&options, args);
+            if (!result.first) {
+                return Help(result.second);
+            }
+        } else {
+            options.tool() = std::move(xcrunBaseName);
```
grp: Modifying the options struct is a little unfortunate. Maybe instead make local variable for each that is assigned differently based on the condition above, then change the code below to use that variable instead of `options`?

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -236,15 +241,25 @@ Version()
...
     Options options;
-    std::pair<bool, std::string> result = libutil::Options::Parse<Options>(&options, args);
-    if (!result.first) {
-        return Help(result.second);
+
+    {
+        std::vector<std::string> args = std::vector<std::string>(argv + 1, argv + argc);
+        std::string xcrunBaseName = libutil::FSUtil::GetBaseName(argv[0]);
+        /* HACK(strager): There must be a better way. */
+        bool isXcrun = xcrunBaseName == "xcrun";
```
grp: From the `xcrun` manpage, the correct check looks to be `!filesystem.isSymbolicLink(executablePath)`.

CMake/Install.cmake
```diff
@@ -12,13 +12,96 @@ set(XCBUILD_INSTALL_DEVELOPER_DIR Developer CACHE STRING "Install path for xcbui
...
+        # be) called.
+        XCBUILD_INSTALL_PROGRAM_ALIAS(TARGET ${TARGET} ALIASED_TARGET xcrun DESTINATION bin)
+      else ()
+        message(AUTHOR_WARNING "XCBUILD_INSTALL_DEVELOPER_DIR cannot install non-executable (${TYPE}) target ${TARGET} with USER_ALIAS")
+      endif ()
+    endforeach ()
+  endif ()
+endfunction ()
+
+# XCBUILD_INSTALL_PROGRAM_ALIAS(ALIASED_TARGET a DESTINATION d TARGET t)
```
grp: Rather than installing an alias for a target, why not just install an alias for a program name? Then, in something like `Aliases/CMakeLists.txt`, just have a list of all of the aliases to install:

    XCBUILD_INSTALL_ALIAS(xcbuild)
    XCBUILD_INSTALL_ALIAS(actool)
    XCBUILD_INSTALL_ALIAS(clang)

I think that'd be simpler overall and would allow installing aliases for tools that are not part of xcbuild itself.

grp:
> Does supporting xcrun without an SDK make sense? (See "HACK/WIP Support xcrun without an SDK." commit.)

I don't think so, but would need more context as to why. Note that `xcrun` *should* fall back to `PATH`, but currently does not. Implementing that might cover this use case.

> I didn't realize xcrun won't necessarily use xcbuild's Developer directory. This makes the behaviour of the tools in ${CMAKE_INSTALL_PREFIX}/bin confusing to me. Is that a known problem or is it expected?

I'd say it's expected, assuming that you're explicitly choosing to install the aliases component.

> Should we just install the user programs directly (duplicating them) instead of making aliases?

I don't think so.

> Should I put the user programs in their own CMake install component? Currently they are in the "libraries" component.

I think that would make sense.

> Will there ever be plans to support xcbuild as a set of APIs instead of a set of executables?

Yes, but probably by including it as a submodule of another project and bundling it with that project. Building or linking against an installed copy never seems like a good idea for C++.

> What's a better name for "normal" installed files? I use "USER".

The ones that go in the developer root? "DEVELOPER"?

strager:
>> Does supporting xcrun without an SDK make sense? (See "HACK/WIP Support xcrun without an SDK." commit.)
>
> I don't think so, but would need more context as to why. Note that xcrun should fall back to PATH, but currently does not. Implementing that might cover this use case.

I was just running

    cd Build
    make install DESTDIR=foo
    env DEVELOPER_ROOT=foo/usr/local/Developer foo/usr/local/bin/xcrun plutil -help

and `xcrun` complained about not finding an SDK. If this isn't supposed to work, I don't know how `xcrun` is supposed to work. =D

>> I didn't realize xcrun won't necessarily use xcbuild's Developer directory. This makes the behaviour of the tools in ${CMAKE_INSTALL_PREFIX}/bin confusing to me. Is that a known problem or is it expected?
>
> I'd say it's expected, assuming that you're explicitly choosing to install the aliases component.
>
>> Should we just install the user programs directly (duplicating them) instead of making aliases?
>
> I don't think so.

Hmm, it sounds like my proposal won't actually help @Ktwu. =S

>> What's a better name for "normal" installed files? I use "USER".
>
> The ones that go in the developer root? "DEVELOPER"?

Heh, no, I meant the files in /usr/local/bin/ and /usr/local/lib/ (as opposed to the Developer dir). By "normal" I meant "what people expect to see after running 'make install' on Linux".

grp:
> and xcrun complained about not finding an SDK. If this isn't supposed to work, I don't know how xcrun is supposed to work. =D

Yeah, it shouldn't error out from that, but it has to right now since it doesn't search `PATH` yet. Once xcrun searches `PATH` it doesn't need an SDK necessarily.

grp:
Actually, I think I was wrong: xcrun should require an SDK. Otherwise, what would the `--show-sdk-path` flag do, for example?

grp:
Searching `PATH` is implemented in #85, but xcrun still needs a developer root, toolchains, and SDK. I believe that's expected.

PR #87, opened 2016-07-15
Support both ImageIO and libpng for reading PNG images.

grp:
- Separate out PNG loading into a separate module.
 - Refactor PNG loading to be less tied to asset compliation.
 - Add an implementation of PNG loading using ImageIO when available.


grp:
I haven't tested this yet. It might make sense to add a unit test that loads various PNG image formats and makes sure they're converted correctly, which would then run against libpng and ImageIO in CI.

grp:
OK, I added a test and made the ImageIO reader work. It turns out CoreGraphics can't handle grayscale + alpha pixel data, so that has to be converted manually.

grp:
It looks like the test failures are because the libpng reader is not premultiplying the alpha as expected.

grp:
@marcsalem, this is ready to go now.

Libraries/graphics/Headers/graphics/PixelFormat.h
```diff
@@ -0,0 +1,149 @@
...
+     */
+    enum class Order {
+        /*
+         * Forward order in memory, i.e. big-endian.
+         */
+        Forward,
+        /*
+         * Reversed order in memory, i.e. little-endian.
+         */
+        Reversed,
```
marcsalem: I find the forward/reverse terminology too vague. I think you should consider enumerating RGBA/BGRA/GA etc.

Pixel Order is also used for left->right and right->left pixel order within a row, so using this for the order of channels in a pixel makes this extra confusing.

grp: Listing out all the possible formats explicitly is impractical, since this class supports `2 * 2 * 7 = 28` formats already and I can imagine many more. That would also mean additional functions to get out the information here as it's much easier to switch on "RGB or Gray" than "RGB, BGR, RGBA, ARGB, BGRA, etc vs GA, G, AG, etc".

CoreGraphics uses "big endian" and "little endian" for this, which helps in that it's obviously referring to the channels, but I find it confusing in that it isn't clear what the base value is that the endianness modifies. This makes it a little more clear in that RGB is the base that this can change the ordering of.

I think the important part is that the pixel format only talks about the format of a specific pixel in the image — not the relationship *between* pixels. So being nested in that makes it more clear that it's just for one pixel.

I'm open to alternatives here on naming, though. This isn't my favorite, just the best I came up with.

marcsalem: You could make the distinction between color and gray in one field, and specific format in another.

grp: That could end up with consistency issues, though, if the fields conflicted.

What's the concern with specifying the format as the combination of different axes?

grp:
@marcsalem, ping?

When this is ready to go, would be great to merge by rebasing the branch onto master so it doesn't squash the history but also stays linear:

    git fetch origin grp/png-imageio
    git rebase master origin/grp/png-imageio --onto master
    git push origin master

(I can also do that if you comment lgtm.)

marcsalem:
Looks good to me, please land it

grp:
Landed on master.

Issue #92, opened 2016-07-19
Swift does not import underlying module


grp:
Split out of #88.

@keith wrote:

> Ok so as far as I can tell, based on your branch, the original issue here is fixed. If this next issue doesn't sound related to you, feel free to close this one and I can open a different one as needed!
> 
> The next issue I've ran into, which I can't yet tell if it's related, is that using xcodebuild it seems like (for better or worse) many imports are implicitly added. For example we have files like this:
> 
>     $ cat Foo.swift
>     extension UIView {
>        ...
>     }
> 
> Where we have no explicit import UIKit, yet these work somehow. I think some of this might have to do with other files that are importing 3rd party dependencies, and maybe those imports are polluting the entire modules namespace. Another similar case we have is:
> 
>     $ cat Foo.swift
>     import UIKit
>     
>     ... NSBundle ...
> 
> Which will also fail to compile because of a missing import to Foundation. Although it seems like in other places, import UIKit will implicitly make the Foundation import. Again I'm not a fan of this behavior, but it has apparently been something we're relying on. I'm currently adding all the missing imports by running xcbuild until it fails, copying the command that fails, running it, fixing the imports, repeat. Also as a note, these don't get surfaced in xcpretty, which would probably make this easier, instead I've been grepping the log for exit-status": 1
> 
> Let me know if you think this is at all related.

@keith wrote:

> Ok so after finding all these missing imports, it seems like that issue only applies to the non-app dependencies of our project. I'm not sure exactly what the difference is there, but I only had to modify imports in our first party dependencies, some from CocoaPods and some just custom frameworks in our main project, but none of our main project files. There could be a difference here since the main project is the only place with a bridging header, which might be a slightly more explicit way of polluting the global namespace 😬

@grp wrote:

> Hm, very interesting! It would be useful to see a comparison of the swiftc invocations between xcbuild and xcodebuild for a target hitting the issue. Hard to imagine what could cause a difference like that, though!

keith:
Ok so interestingly enough, I think I figured out a little bit more about this. And both are intricacies with CocoaPods and how it generates dynamic frameworks.

So there are a few parts to this. First CocoaPods includes a `pch` file via `GCC_PREFIX_HEADER`. This imports `UIKit` globally..

Second in the generated umbrella headers, CocoaPods also imports `UIKit`. So if you remove the `GCC_PREFIX_HEADER` build setting via your Podfile, you still have this imported everywhere.

[...]

PR #93, opened 2016-07-19
Add support for `MODULEMAP_FILE` build setting. Fixes #88.

### LINKED ISSUE #88: Add support for VFS-pointed module maps (was: can't find module dependency)
While trying to setup the Lyft project to use `xcbuild`, I ran into some issues with dependencies. We use CocoaPods to integrated 3rd party, and some first party development dependencies. The structure looks something like this:

```
Main Project
  -> Alamofire
  -> Internal Library
       -> zipzap
  ...
```

In this simplified example, the issue we're running into is that `zipzap` cannot be found. This leads to errors in the `Internal Library`, and stops us from getting anywhere else.

I've attempted to reproduce the same issue in a sample project:

[test-xcbuild4.zip](https://github.com/facebook/xcbuild/files/368231/test-xcbuild4.zip)

In this project if you run (with or without specifying `-executor ninja`):

```
$ xcbuild -executor ninja -workspace test-xcbuild4 -scheme test-xcbuild4
```

You will see that it fails to build with the error:

```
{
  "kind": "finished",
  "name": "compile",
  "pid": 48974,
  "output": "blah blah/AppDelegate.swift:10:8: error: no such module 'zipzap'\nimport zipzap\n       ^\n",
  "exit-status": 1
}
```

(here's the [entire build log](https://github.com/facebook/xcbuild/files/368236/buildlog.zip))

But if you build from within Xcode, everything compiles correctly.

I would love some advice for where to start when I'm facing errors like this, as I'd love to be able to set our project up to benefit from the speed of Ninja! ⚡ 

Thanks!

grp:
Implements the module map features needed for #88. They're part of the grab bag of build things in `ProductTypeResolver`; that probably should be split up in a later PR. For now, this fixes that task and should be a big help for any Swift projects using CocoaPods.


facebook-github-bot-3:

grp:
I added an additional set of commits on top to fix #92 as well. This is a pretty big change at this point, and I don't have great test cases for it (just a few sample projects I made), but if works for @keith I think it's likely pretty solid.

Ktwu:
I'm working through this now.  For lack of a better strategy, I'm going through your commits one by one.

"Support auxiliary files with dynamic content from the file system" LGTM.  What's an auxiliary file?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+    std::transform(moduleName.begin(), moduleName.end(), std::back_inserter(lowerModuleName), ::tolower);
+
+    ext::optional<std::string> umbrellaHeader;
+    for (pbxproj::PBX::BuildPhase::shared_ptr const &buildPhase : target->buildPhases()) {
+        if (buildPhase->type() != pbxproj::PBX::BuildPhase::Type::Headers) {
+            continue;
+        }
+
+        for (pbxproj::PBX::BuildFile::shared_ptr const &buildFile : buildPhase->files()) {
+            if (buildFile->fileRef() == nullptr) {
```
Ktwu: A phase's build file list contains all the files that phase uses, right?  When would this ref be null?

grp:
An auxiliary file is a file generated by the build system, rather than coming from the source code or being generated by a rule. For example, script build phases have an inline script in the project; those require an auxiliary file to write out the shell script to the filesystem to run it. Or header maps, which are completely generated by the build system but are needed to compile.

(If that makes sense, I'll add it to `AuxiliaryFile`'s header...)

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
     return true;
 }
 
+static ext::optional<std::string>
+FindUmbrellaHeaderName(pbxsetting::Environment const &environment, std::string const &moduleName, pbxproj::PBX::Target::shared_ptr const &target)
```
Ktwu: An umbrella header example would be like if I had a project named "Foo", and also had a header named "Foo.h", right?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+{
+    std::string lowerModuleName;
+    std::transform(moduleName.begin(), moduleName.end(), std::back_inserter(lowerModuleName), ::tolower);
```
Ktwu: Why bother transforming the module name?  If I have a module named "Foo", I shouldn't pick up on a file named "foo.h" or "fOo.h" for my header name.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Would it be reasonable to write a small unit test for something like this function?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+
+    return ext::nullopt;
+}
+
+static bool
+ResolveModuleMap(Phase::Environment const &phaseEnvironment, Phase::Context *phaseContext, pbxsetting::Environment const &environment)
+{
+    std::string const &workingDirectory = phaseContext->toolContext().workingDirectory();
+    // TODO: what about non-framework product types?
+    std::string finalModuleMapRoot = environment.resolve("TARGET_BUILD_DIR") + "/" + environment.resolve("CONTENTS_FOLDER_PATH") + "/" + "Modules";
```
Ktwu: Aww, there's no constant for the separator?  It'd be nice if there was a helper function like Python's os.path.join...

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: The phase and file objects here are directly deserialized from the project. Malformed projects might have broken references all throughout, and many projects are malformed after a few rounds of merge conflicts...

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: Yeah, exactly, although based on the module name rather than the project name.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: HFS+ is case-insensitive by default, so people get this wrong all the time. For example Pop for a long time had the target name `pop` but the header was called `POP.h`, since it happened to work at the time.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+                }
+            }
+        }
+    }
+
+    return ext::nullopt;
+}
+
+static bool
+ResolveModuleMap(Phase::Environment const &phaseEnvironment, Phase::Context *phaseContext, pbxsetting::Environment const &environment)
```
Ktwu: Remind me of your reasoning behing using a pointer vs. reference -- you reserve pointers for large objects, correct?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: I agree on both counts, fixing that is #6.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: Pointers are for anything mutable, so at the call site you can see it can be mutated (i.e. the `&` for the pointer) rather than being identical to passing by value. In this case `Phase::Context` holds a `Tool::Context` which has a mutable `invocations` vector that this adds to, so it's a pointer to note that it's being mutated.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+    if (moduleMapAuxiliaryFile) {
+        /* Define source module map. */
+        // TODO: it would be nicer to attach this to the copy invocation created below
+        Tool::Invocation invocation;
+        invocation.auxiliaryFiles().push_back(*moduleMapAuxiliaryFile);
+        phaseContext->toolContext().invocations().push_back(invocation);
+
+        /* Copy in the module map as part of the build. */
+        if (Tool::CopyResolver const *copyResolver = phaseContext->copyResolver(phaseEnvironment)) {
+            copyResolver->resolve(&phaseContext->toolContext(), environment, { intermediateModuleMapPath }, finalModuleMapRoot, "Ditto");
```
Ktwu: "Ditto"?

grp: The example log I was basing this on said "Ditto" in the log message. Not sure why, since it's not actually using the `/usr/bin/ditto` tool, but I guess since it's just a simple copy.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -268,6 +269,135 @@ ResolveFrameworkStructure(Phase::Environment const &phaseEnvironment, Phase::Con
...
+
+        /* Copy in the module map as part of the build. */
+        if (Tool::CopyResolver const *copyResolver = phaseContext->copyResolver(phaseEnvironment)) {
+            copyResolver->resolve(&phaseContext->toolContext(), environment, { intermediatePrivateModuleMapPath }, finalModuleMapRoot, "Ditto");
+        } else {
+            fprintf(stderr, "warning: failed to get copy tool for module map\n");
+        }
+    }
+
+    return true;
```
Ktwu: This should probably return false at some point, but I only see this 'true' return statement.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
```diff
@@ -335,6 +465,15 @@ resolve(Phase::Environment const &phaseEnvironment, Phase::Context *phaseContext
     }
 
     /*
+     * Create the product's module maps, if required.
+     */
+    if (pbxsetting::Type::ParseBoolean(environment.resolve("DEFINES_MODULE"))) {
```
Ktwu: Dumb question:  What's a module map?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: It's mostly returning `bool` to match the other types in `pbxbuild::Phase` that return `bool`, not sure it necessarily can fail itself right now.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: @compnerd could probably answer better than me. It's it's a file that does some stuff I don't really understand but (in this case) is used to to make groups of Objective-C headers into Swift modules.

modocache: http://clang.llvm.org/docs/Modules.html#module-maps has some documentation -- but even after reading it several times, I still find module maps to be pretty confusing 😅

Ktwu:
Yeah, that explanation about auxiliary files makes plenty sense, thanks.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: xcbuild is also intended to run on Linux, which does have a case sensitive filesystem.  My own experiments building on Linux have tripped over stupid header misnames, which I've had to fix.

Since this is something Linux builders would have to worry about anyway, can you make this strict?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Cool.  Is there precendent for adding something like TODO(#6) in xcbuild?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Ok.  What's the downside of using const?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: The documentation helps clear things up for me, thanks.  I now the difference between #include and #import's behaviors!

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Hmm...what about in the case of your comment:  "// TODO: what about non-framework product types?"

Or what about if invocation.auxiliaryFiles() never has a module map added to it?

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
```diff
@@ -17,6 +17,7 @@
 #include <algorithm>
 
 namespace Tool = pbxbuild::Tool;
+namespace Phase = pbxbuild::Phase;
 using libutil::FSUtil;
```
Ktwu: nit: newline pls

Ktwu: nit: newline pls

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
```diff
@@ -25,13 +26,15 @@ CopyResolver(pbxspec::PBX::Tool::shared_ptr const &tool) :
...
-resolve(
+static void
+ResolveInternal(
+    pbxspec::PBX::Tool::shared_ptr const &tool,
     Tool::Context *toolContext,
     pbxsetting::Environment const &baseEnvironment,
-    std::vector<Phase::File> const &inputs,
+    std::vector<std::string> const &inputPaths,
+    ext::optional<std::vector<Phase::File>> const &inputs,
     std::string const &outputDirectory,
```
Ktwu: Completely random aside:  I've seen in Buck that strings that represent paths are passed around in their own object type called a SourcePath.  Is there anything similar in xcbuild?

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
```diff
@@ -95,6 +100,47 @@ resolve(
...
+void Tool::CopyResolver::
+resolve(
+    Tool::Context *toolContext,
+    pbxsetting::Environment const &baseEnvironment,
+    std::vector<Phase::File> const &inputs,
+    std::string const &outputDirectory,
+    std::string const &logMessageTitle) const
+{
+    std::vector<std::string> inputPaths;
+    std::transform(inputs.begin(), inputs.end(), std::back_inserter(inputPaths), [&](Phase::File const &input) -> std::string {
```
Ktwu: Why not use foreach syntax?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: Can't forget `@import`! So many ways to import things.

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
grp: where?

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
grp: Nope. That's filed as #6.

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
grp: I think I copied these lines from somewhere else that used this syntax, but not sure why.

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
Ktwu: Between lines 20 and 21.

grp: Hmm, maybe I remember wrong but I don't think the rest of xcbuild does that?

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Ah, ok.  Nevermind then.

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
Ktwu: Oh, ok, if it's not the project's style then nvm.

Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
Ktwu: Ok.  I find foreach more palatable, but not a big deal.

Libraries/pbxbuild/Headers/pbxbuild/Tool/CopyResolver.h
```diff
@@ -34,6 +34,13 @@ class CopyResolver {
         std::string const &outputDirectory,
```
Ktwu: This commit LGTM.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -598,224 +670,185 @@ NinjaDependencyInfoExecutable()
...
-                }
-            }
+    /*
+     * Build up the command to create the auxiliary file.
+     */
+    std::string exec;
+    if (auxiliaryFile.contentsData()) {
+        // FIXME: use base64 directly, rather than through plist
+        auto data = plist::Data::New(*auxiliaryFile.contentsData());
+        exec = "echo " + data->base64Value() + " | base64 --decode > " + Escape::Shell(auxiliaryFile.path());
```
Ktwu: :'(

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
compnerd: A module map is merely a mapping used to construct the "module".  The module presents a logical "module" which provides a certain set of interfaces (macros, functions, structures, and in the case of C++, classes and templates) which can then be imported as a single unit.  This is also the underlying mechanism behind the interoperability between swift and C.

facebook-github-bot-3:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -598,224 +670,185 @@ NinjaDependencyInfoExecutable()
...
-    if (!WriteNinja(filesystem, writer, path)) {
-        fprintf(stderr, "error: unable to write target ninja: %s\n", path.c_str());
-        return false;
+    std::vector<ninja::Value> orderDependencies;
+    for (std::string const &orderDependency : invocation.orderDependencies()) {
+        orderDependencies.push_back(ninja::Value::String(orderDependency));
     }
 
+    /*
+     * All invocations depend on the target containing them beginning.
```
Ktwu: I don't understand this comment.  What do you mean by "containing them beginning"?

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -598,224 +670,185 @@ NinjaDependencyInfoExecutable()
...
 bool NinjaExecutor::
-buildTargetAuxiliaryFiles(
-    Filesystem *filesystem,
+buildAuxiliaryFile(
     ninja::Writer *writer,
-    pbxproj::PBX::Target::shared_ptr const &target,
-    pbxbuild::Target::Environment const &targetEnvironment,
-    std::vector<pbxbuild::Tool::Invocation> const &invocations)
+    pbxbuild::Tool::Invocation::AuxiliaryFile const &auxiliaryFile,
+    std::string const &after)
```
Ktwu: "after" means "Please perform this step after this other rule" right?

Libraries/xcexecution/Headers/xcexecution/NinjaExecutor.h
```diff
@@ -11,6 +11,7 @@
 #define __xcexecution_NinjaExecutor_h
```
Ktwu: Overall this commit LGTM.  I know little about ninja, though, so my LGTM is more like "I trust you".

Libraries/xcexecution/Sources/NinjaExecutor.cpp
grp: Yep.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
grp: Ninja models dependencies at the file level, but targets often have dependencies on every file for another target being completed. To model that in Ninja, xcbuild sets up special rules for each target:

 - A rule that marks the target starting, and depends on each of its target dependencies' finish rule (see below). Each file rule within the target depends on this rule in order to build after any dependent targets finish.
 - A rule that marks the target finishing, and depends on each rule in the target finishing. Dependent targets begin rule (see above) depend on this in order to build after it.

This comment is noting the second part of the begin rule above — that all files within the target depend on the target's begin rule.

grp: There's also a third special rule added by this PR, that goes between a target's begin rule and the files in the target. That special rule is used as a barrier between writing out auxiliary files and beginning the target's file rules, in a similar way to the barriers between targets. I forget if that's added in the commit you're looking at here or the other one. :(

Libraries/pbxbuild/Headers/pbxbuild/Tool/Invocation.h
```diff
@@ -19,26 +19,61 @@ namespace Tool {
 class Invocation {
 public:
     class AuxiliaryFile {
+    public:
```
Ktwu: This commit LGTM.  I'm guessing I'll see this in use in your other commits.

Libraries/pbxbuild/Headers/pbxbuild/Tool/Invocation.h
```diff
@@ -60,8 +60,8 @@ class Invocation {
         std::vector<Chunk> _chunks;
```
Ktwu: Nice commit.  LGTM.

Libraries/pbxbuild/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ add_library(pbxbuild SHARED
             Sources/Phase/CopyFilesResolver.cpp
```
Ktwu: Looks like you just moved code around.  LGTM.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
Ktwu: Don't worry, it's in this commit's message.

Libraries/pbxbuild/Headers/pbxbuild/Tool/ModuleMapInfo.h
```diff
@@ -0,0 +1,92 @@
...
+    /*
+     * The public module map.
+     */
+    ext::optional<Entry> const &moduleMap() const
+    { return _moduleMap; }
+    ext::optional<Entry> &moduleMap()
+    { return _moduleMap; }
+
+    /*
+     * The private module map.
```
Ktwu: If public modules are intended for exposing symbols to other modules in an organized fashion...what are private module maps for?  The owning module?

Libraries/pbxbuild/Sources/Phase/ModuleMapResolver.cpp
```diff
@@ -29,143 +30,64 @@ Phase::ModuleMapResolver::
...
-    std::string finalModuleMapRoot = environment.resolve("TARGET_BUILD_DIR") + "/" + environment.resolve("CONTENTS_FOLDER_PATH") + "/" + "Modules";
+    Tool::CopyResolver const *copyResolver = phaseContext->copyResolver(phaseEnvironment);
+    if (copyResolver == nullptr) {
+        fprintf(stderr, "warning: failed to get copy tool for module map\n");
+        return false;
+    }
 
     /*
-     * Handle the standard module map, either copied or generated.
+     * Create the module map info. Note this is required even if a module is not defined, in order to
```
Ktwu: This reads as "a module map is needed for builds that don't define a module."  Seems weird.

Ktwu: Ah, nevermind, I see this is 'module map info'.

Libraries/pbxbuild/CMakeLists.txt
```diff
@@ -21,13 +21,15 @@ add_library(pbxbuild SHARED
             Sources/Tool/CompilationInfo.cpp
```
Ktwu: Seems sensible.  LGTM.

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
```diff
@@ -232,6 +232,73 @@ AppendObjcHeader(
...
+    std::string const &headerName)
+{
+    /* No module map means no underlying module to import. */
+    ext::optional<Tool::ModuleMapInfo::Entry> const &moduleMap = toolContext->moduleMapInfo().moduleMap();
+    if (!moduleMap) {
+        return;
+    }
+
+    /*
+     * Create a module map to import the underlying module. However, exclude the Swift-generated
```
Ktwu: This entire commit is so Swift libraries can use the ObjC version of their APIs internally, right?

Given your comment about avoiding importing the module's own structures, I'm guessing that's a bad thing; why is it not necessary?

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
```diff
@@ -232,6 +232,73 @@ AppendObjcHeader(
...
+    auxiliaryFiles->push_back(unextendedModuleMapFile);
+
+    /*
+     * Create a VFS overlay to load the above module map. This replaces the module map in the product,
+     * which does not contain the extra Swift rule, and may not even exist yet (depending on ordering).
+     */
+    std::string unextendedModuleOverlayPath = environment.resolve("TARGET_TEMP_DIR") + "/" + "unextended-module-overlay.yaml";
+
+    std::string unextendedModuleOverlay;
+    unextendedModuleOverlay += "{\n";
```
Ktwu: Isn't there a JSON writer you could use instead?

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
```diff
@@ -232,6 +232,73 @@ AppendObjcHeader(
     }
 }
 
+static void
```
Ktwu: LGTM.

Ktwu:
You mentioned that if this patch works for @keith then it ought to be reasonable.  Have they tested this yet?

Also, do you think you could upload the projects you used for testing?  At the very least I imagine xcbuild ought to have a few sanity projects to build in the future.

Libraries/pbxbuild/Headers/pbxbuild/Tool/ModuleMapInfo.h
grp: If you build a framework for use internally, it can also include a `PrivateHeaders/` directory. The `module.private.modulemap` would probably reference headers in that.

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
grp: This is YAML, not JSON, although I think it's valid JSON too so I suppose it could use the `plist` API to write it out. But it's simple enough this seemed easiest?

grp:
I sadly don't have a good test case set for xcbuild, but that sounds like a good idea! Maybe a separate repo with just test cases to try building.

Not sure if @keith has tried the latest version — if not, that'd be awesome to make sure this actually fixes the issue before landing!

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
grp: The goal is two parts:

 1. Give Swift code in a module access to the Objective-C headers from the same module. This requires a module map for the Objective-C code so Swift can know how to load it.
 2. Automatically load the module's Objective-C half from Swift. This is accomplished with the `-import-underlying-module` flag added below.

The issue is that compiling Swift code also generates an Objective-C header, to let other Objective-C code use the classes, methods, etc from Swift. What this is trying to avoid is circularly loading that header back into Swift as if it was part of the manually written Objective-C module.

For example, let's say you have an Objective-C class `GRPOcean`, and a Swift class `Continent`, both in the same module (i.e. same target). The Swift class should be able to see the Objective-C class without importing it explicitly, and the Objective-C class can see the Swift class by importing the generated Swift header. But we don't want the Swift class to import the generated Swift header, since that defines the same thing the Swift code does!

keith:
Sorry I haven't tried to do this again yet!

Libraries/pbxbuild/Sources/Tool/SwiftResolver.cpp
Ktwu: Yeah, this is fine, my comment is a nit.  If you revisit this code, I recommend at the very least using a multiline string.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Ping on this question.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Ping on this question.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
Ktwu: Ping.

Ktwu:
Once you address my last few noob questions and after @keith tests, then LGTM.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: In these functions, the return value is more like "success" than "did it do anything". So in those cases it would still return true since nothing went *wrong*, it just didn't do anything. It might make sense to return `false` if it can't find the copy tool, but that's a wider refactor since I haven't figured out the best way to report missing tools in general yet.

Libraries/pbxbuild/Sources/Phase/ProductTypeResolver.cpp
grp: `const` for what? `Phase::Context *phaseContext` can't be `const` because it's mutated (via `phaseContext->toolContext().invocations()`).

Ktwu: Oops, missing context:  instead of using references to represent immutable objects (since you're using pointers for mutable stuff), why not just use const pointers?  I'm just curious.

grp: Ah! I only see one real choice: passing a copy separate from the caller (i.e. pass-by-value), or passing a pointer to change the caller's value (i.e. pass-by-reference). Then the question is what the best way is to implement each of those.

For passing a copy, a const reference has the same semantics of passing a copy — but without the performance cost of actually copying. So a const reference is just an optimization or implementation detail of the semantics of pass by value. For passing a pointer, both options have the same performance. So it falls back to the argument about which is most obvious to the caller about what might be mutated.

The other two cases, const pointer and non-const reference, end up not fitting into either group, so I don't use them much. Unused features are what make C++ fun, I guess.

keith:
It looks like with the newest version of this branch, I'm ending up with some new errors:

```
{
  "kind": "finished",
  "name": "compile",
  "pid": 15676,
  "output": "\/Users\/ksmiley\/Library\/Developer\/Xcode\/DerivedData\/Example-agquzgvxfwspjpgpojvjjsmguhuv\/Build\/Intermediates\/Pods.build\/Debug-iphonesimulator\/Alamofire.build\/unextended-module.modulemap:2:19: error: umbrella header 'Alamofire-umbrella.h' not found\n  umbrella header \"Alamofire-umbrella.h\"\n                  ^\n<unknown>:0: error: could not build Objective-C module 'Alamofire'\n",
  "exit-status": 1
}
```

I'm now getting these errors for our framework dependencies. But then when I check the build directory for the expected headers, they seem to be there (at least in the built products):

```
(dev/xcbuild)[grp/modulemap-file] % find ~/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build -name "*-umbrella.h"
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/Alamofire/Alamofire.framework/Headers/Alamofire-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/JGProgressHUD/JGProgressHUD.framework/Headers/JGProgressHUD-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/LambdaKit/LambdaKit.framework/Headers/LambdaKit-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/libPhoneNumber-iOS/libPhoneNumber_iOS.framework/Headers/libPhoneNumber-iOS-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/LyftKit/LyftKit.framework/Headers/LyftKit-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/LyftUI/LyftUI.framework/Headers/LyftUI-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/ModelMapper/Mapper.framework/Headers/ModelMapper-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/SnapKit/SnapKit.framework/Headers/SnapKit-umbrella.h
/Users/ksmiley/Library/Developer/Xcode/DerivedData/Example-agquzgvxfwspjpgpojvjjsmguhuv/Build/Products/Debug-iphonesimulator/zipzap/zipzap.framework/Headers/zipzap-umbrella.h
```

Interestingly after running the xcbuild command a few more times, the issue seemed to clear up...

keith:
Actually these _might_ not have cleared up immediately, as I just saw a case where we got a `"exit-status": 1` from one of these commands, but xcbuild exited successfully.

keith:
I didn't see any errors related to the problem in #93, but that might be misleading because of these other errors.

grp:
That looks like progress: it's trying to automatically import the underlying Objective-C module, but fails.

> Interestingly after running the xcbuild command a few more times, the issue seemed to clear up...

It looks like this is because there's no ordering on copying the umbrella header versus trying to use the module map that references is. Was able to reproduce locally with Alamofire, will investigate what that ordering should look like.

> Actually these might not have cleared up immediately, as I just saw a case where we got a "exit-status": 1 from one of these commands, but xcbuild exited successfully.

This is quite worrying. Were you using the Ninja executor or the built-in one when you saw this? Either way, it should be handling command failures correctly and stopping.

keith:
I saw this with the ninja executor. I'll try and debug it next time I spot it.

grp:
It turns out the fix isn't more ordering, it's to use a VFS to emulate having already copied the headers. I have a partial fix for that, but generating the VFS is a little more complicated than I hoped since it's shared between builds of all targets in a project.

grp:
I landed this now so it doesn't conflict with rebases, even though the VFS part still isn't done. I'll put up a WIP for that now though.

grp:
Remaining work is in 948f5b45d33fb76a4cb2b3d8906d24dc6114685e (on `grp/vfs-overlay`) for now.

PR #94, opened 2016-07-21
Support compiling app icon sets.

grp:
Supports copying the images and writing the icon metadata to info
plist additions. Current implementation has two limitations:
    
 - All idioms are copied for all target platforms. For example, watch images are copied for asset compilation targeting a phone.
 - The size of images is not verified against the specified dimensions.


facebook-github-bot-8:

marcsalem:
It would be awesome if there was a unit test, but that can be a follow up task for now. Thanks!

PR #95, opened 2016-07-21
Use optionals in parsing command line arguments.

grp:
Instead of checking if strings are empty, use optionals to store if
the arguments were passed at all. Additionally, add a few utility
methods to the option parser to simplify various tool option classes.

### DIFF
--- a/Libraries/libutil/Sources/Options.cpp
+++ b/Libraries/libutil/Sources/Options.cpp
@@ -13,15 +13,18 @@
 
 using libutil::Options;
 
+namespace libutil {
+
+template<>
 std::pair<bool, std::string> Options::
-NextString(std::string *result, std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it, bool allowDuplicate)
+Next<std::string>(ext::optional<std::string> *result, std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it, bool allowDuplicate)
 {
     std::string const &arg = **it;
-    ++(*it);
+    if (++*it != args.end()) {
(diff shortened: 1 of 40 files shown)


facebook-github-bot-0:

modocache:
@grp It looks like your previous CLA is no longer "active". I think you should be able to renew your agreement (and shut these bots up) by visiting https://code.facebook.com/cla.

grp:
I sent a new one to @caabernathy last week (the online form errors out), assuming it just hasn't been added into the system yet.

modocache:
Awesome, thanks! Sorry for the noise :)

grp:
@Ktwu, not sure why, but your name doesn't seem to load in the assignees dropdown. Maybe you need to join the @facebook org?

Libraries/libutil/Sources/Options.cpp
```diff
@@ -31,42 +34,93 @@ NextString(std::string *result, std::vector<std::string> const &args, std::vecto
...
-        if (str == "YES") {
-            *result = true;
-        } else if (str == "NO") {
-            *result = false;
+    std::string const &arg = **it;
+    if (++*it != args.end()) {
+        std::string const &value = **it;
+
+        if (!*result || allowDuplicate) {
+            if (value == "YES") {
```
marcsalem: This looks like "no" and "yes" will fail. Is it really expected to be case sensitive?

marcsalem:
Looks good to me.

Ktwu:
@grp, I've added myself to the @facebook org, so hopefully I show up in the future.

PR #97, opened 2016-07-27
Minor cleanups for asset compilation.

grp:
- Move asset compilation output into the `acdriver::Compile` namespace.
 - Centralize various conversion functions so they can be shared between asset types.

This is a prerequisite for a later PR that adds support for compiling launch images.


marcsalem:
Looks like a straight forward refactor. I'll land it once tests pass.

PR #99, opened 2016-07-27
No error if there is no dependancy info to write

marcsalem:
We currently fail if actool is called with --export-dependency-info specified.


grp:
The issue with dependency info AFAIK is that nothing populates the dependency info fields of the output in the various asset types' compilation. If that was populated, it should be fine if it turned out empty and an empty file was written out?

marcsalem:
In the filesystem abstraction, this is written with std::fwrite(), which returns the number of items written, which is zero. So we get an error when writing and empty vector.

grp:
Ah, I must have broken that in f0e3ace4a3eea82be5dc31ed9c8cb085c5b0842b. I'll push a fix for `DefaultFilesystem`.

grp:
Fixed by b8919d42485e02294434596bb87378ddeef63243.

PR #105, opened 2016-08-01
WIP Use lzfse as a portable alternative to libcompression.

grp:
WIP pending https://github.com/lzfse/lzfse/pull/25 and https://github.com/lzfse/lzfse/pull/26.

Allows decompressing LZFSE and LZVN compression on Linux.

Should the lzfse submodule be required? That would make the default build require git submodules be cloned, which isn't necessarily right now. Alternatively, could make either libcompression or lzfse required, or leave them both as optional.

### DIFF
--- a/Libraries/libcar/CMakeLists.txt
+++ b/Libraries/libcar/CMakeLists.txt
@@ -20,12 +20,24 @@ find_package(ZLIB REQUIRED)
 target_include_directories(car PRIVATE "${ZLIB_INCLUDE_DIR}")
 target_link_libraries(car PRIVATE ${ZLIB_LIBRARIES})
 
-find_library(COMPRESSION compression)
-if ("${COMPRESSION}" STREQUAL "COMPRESSION-NOTFOUND")
-  set(COMPRESSION "")
+set(LZFSE_SOURCE_DIR "${PROJECT_SOURCE_DIR}/ThirdParty/lzfse")
+if (IS_DIRECTORY "${LZFSE_SOURCE_DIR}")
+  set(LZFSE_BINARY_DIR "${PROJECT_BINARY_DIR}/lzfse") 
+  set(LZFSE_BUNDLE_MODE FALSE)
+  add_subdirectory("${LZFSE_SOURCE_DIR}" "${LZFSE_BINARY_DIR}" EXCLUDE_FROM_ALL)
+
+  target_compile_definitions(car PRIVATE HAVE_LZFSE=1)
+  target_link_libraries(car PRIVATE lzfse)
+  target_include_directories(car PRIVATE "${LZFSE_SOURCE_DIR}/src")
 endif ()
 
-target_link_libraries(car PUBLIC ext bom ${COMPRESSION})
+find_library(COMPRESSION compression)
+if (NOT "${COMPRESSION}" STREQUAL "COMPRESSION-NOTFOUND")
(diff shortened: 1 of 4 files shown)


facebook-github-bot-9:

facebook-github-bot-8:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

marcsalem:
I would be in favor of a submodule in this case since the library is much less mature (changing faster than libpng) and less common to see it supported already in a distro.

But use HTTPS not SSH.

Libraries/libcar/CMakeLists.txt
```diff
@@ -20,12 +20,24 @@ find_package(ZLIB REQUIRED)
...
+  set(LZFSE_BINARY_DIR "${PROJECT_BINARY_DIR}/lzfse") 
+  set(LZFSE_BUNDLE_MODE FALSE)
+  add_subdirectory("${LZFSE_SOURCE_DIR}" "${LZFSE_BINARY_DIR}" EXCLUDE_FROM_ALL)
+
+  target_compile_definitions(car PRIVATE HAVE_LZFSE=1)
+  target_link_libraries(car PRIVATE lzfse)
+  target_include_directories(car PRIVATE "${LZFSE_SOURCE_DIR}/src")
+endif ()
+
 find_library(COMPRESSION compression)
```
strager: Don't bother searching for compression if LZFSE is found.

strager:
> Should the lzfse submodule be required? That would make the default build require git submodules be cloned, which isn't necessarily right now. Alternatively, could make either libcompression or lzfse required, or leave them both as optional.

The "CMake way" is to:

* Expose an option called `XCBUILD_USE_LZFSE` (or something) to enable/disable lzfse support,
* Create a CMake module called `FindLzfse.cmake`,
* If `XCBUILD_USE_LZFSE` is truthy:
 * Use your module: `find_package(Lzfse)`,
 * Fall back to compiling the source version if `LZFSE_FOUND` is falsy and the source version exists, and
 * Error out if `XCBUILD_USE_LZFSE` was explicitly set and `LZFSE_FOUND` is falsy.

(Things may be a bit different lzfse wants to use pkgconfig or CPack.)

This gives the user the power to configure things how they want.

* If the user doesn't care about lzfse support, they don't need to do anything. Things will configure whether or not lzfse is installed or its source is checked out.
* If lzfse ever becomes stable, you don't need to touch xcbuild's CMakeLists.txt in order for users to benefit from their system lzfse.
* If the system lzfse is incompatible with xcbuild, `-DLZFSE_FOUND=YES` falls back to the in-source version.
* If the user has a custom lzfse build, they can configure flags with `-DLZFSE_LIBRARIES=/path/to/lzfse.a` and such. (This is how Facebook would configure xcbuild.)
* If the user doesn't want lzfse support, `-DXCBUILD_USE_LZFSE=NO` disables everything.

strager:
TL;DR provide a way for Facebook to configure where lzfse's libs/includes or project source is.

grp:
@strager Sounds good. Is it OK for Facebook if I make this require my fork of lzfse (https://github.com/grp/lzfse) instead, to access the LZVN APIs? Or would it be preferable to reach into the lzfse internals from upstream inside xcbuild?

strager:
> Is it OK for Facebook if I make this require https://github.com/grp/lzfse instead, to access the LZVN APIs?

Yeah, that's fine.

PR #107, opened 2016-08-12
Make reading and writing CAR files faster

marcsalem:
Avoid memmoves in the libbom by optionally preallocating space for the indexes
Allow raw reads and writes of rendition key-value pairs
Allow specifying the key format to the writer

### DIFF
--- a/Libraries/libbom/Headers/bom/bom.h
+++ b/Libraries/libbom/Headers/bom/bom.h
@@ -42,6 +42,9 @@ struct bom_context *
 bom_alloc_empty(struct bom_context_memory memory);
 
 struct bom_context *
+bom_alloc_size(struct bom_context_memory memory, uint32_t index_count);
+
+struct bom_context *
 bom_alloc_load(struct bom_context_memory memory);
 
 struct bom_context_memory const *
@@ -88,6 +91,9 @@ struct bom_tree_context;
 
 struct bom_tree_context *
 bom_tree_alloc_empty(struct bom_context *context, const char *variable_name);
+
+struct bom_tree_context *
+bom_tree_alloc_size(struct bom_context *context, const char *variable_name, uint32_t index_count);
 
 struct bom_tree_context *
 bom_tree_alloc_load(struct bom_context *context, const char *variable_name);
--- a/Libraries/libbom/Headers/bom/bom.h
+++ b/Libraries/libbom/Headers/bom/bom.h
@@ -39,9 +39,6 @@ bom_context_memory_file(const char *fn, bool writeable, size_t minimum_size);
 struct bom_context;
 
 struct bom_context *
-bom_alloc_empty(struct bom_context_memory memory);
-
-struct bom_context *
 bom_alloc_size(struct bom_context_memory memory, uint32_t index_count);
 
 struct bom_context *
@@ -88,9 +85,6 @@ bom_variable_add(struct bom_context *context, const char *name, int data_index);
 /* Tree */
 
 struct bom_tree_context;
-
-struct bom_tree_context *
-bom_tree_alloc_empty(struct bom_context *context, const char *variable_name);
 
 struct bom_tree_context *
 bom_tree_alloc_size(struct bom_context *context, const char *variable_name, uint32_t index_count);
--- a/Libraries/libbom/Headers/bom/bom.h
+++ b/Libraries/libbom/Headers/bom/bom.h
@@ -39,7 +39,10 @@ bom_context_memory_file(const char *fn, bool writeable, size_t minimum_size);
 struct bom_context;
 
 struct bom_context *
-bom_alloc_size(struct bom_context_memory memory, uint32_t index_count);
+bom_alloc_empty(struct bom_context_memory memory);
+
+struct bom_context *
+bom_alloc_empty2(struct bom_context_memory memory, uint32_t index_count);
 
 struct bom_context *
 bom_alloc_load(struct bom_context_memory memory);
@@ -87,7 +90,10 @@ bom_variable_add(struct bom_context *context, const char *name, int data_index);
 struct bom_tree_context;
 
 struct bom_tree_context *
-bom_tree_alloc_size(struct bom_context *context, const char *variable_name, uint32_t index_count);
+bom_tree_alloc_empty(struct bom_context *context, const char *variable_name);
+
+struct bom_tree_context *
+bom_tree_alloc_empty2(struct bom_context *context, const char *variable_name, uint32_t index_count);
 
 struct bom_tree_context *
 bom_tree_alloc_load(struct bom_context *context, const char *variable_name);
--- a/Libraries/libbom/Headers/bom/bom.h
+++ b/Libraries/libbom/Headers/bom/bom.h
@@ -44,6 +44,9 @@ bom_alloc_empty(struct bom_context_memory memory);
 struct bom_context *
 bom_alloc_empty2(struct bom_context_memory memory, uint32_t index_count);
 
+void
+bom_alloc_indexes(struct bom_context *context, uint32_t index_count);
+
 struct bom_context *
 bom_alloc_load(struct bom_context_memory memory);
 
--- a/Libraries/libbom/Sources/bom.c
+++ b/Libraries/libbom/Sources/bom.c
@@ -40,12 +40,6 @@ _bom_alloc(struct bom_context_memory memory)
 }
 
 struct bom_context *
-bom_alloc_empty(struct bom_context_memory memory)
-{
-  return bom_alloc_size(memory, 2);
-}
-
-struct bom_context *
 bom_alloc_size(struct bom_context_memory memory, uint32_t index_count)
 {
     struct bom_context *context = _bom_alloc(memory);
--- a/Libraries/libbom/Sources/bom.c
+++ b/Libraries/libbom/Sources/bom.c
@@ -246,17 +246,12 @@ bom_index_add(struct bom_context *context, const void *data, size_t data_len)
     struct bom_index_header *index_header = (struct bom_index_header *)((void *)header + ntohl(header->index_offset));
 
     /* Insert index at the end of the list. */
-    bool resize;
-    size_t new_index_length = sizeof(struct bom_index_header) + sizeof(struct bom_index) * (ntohl(index_header->count) + 1);
-    if (new_index_length < ntohl(header->index_length) - (sizeof(struct bom_index) * 2) ) {
-        resize = false;
-    } else {
-        resize = true;
-    }
     uint32_t index_point = ntohl(header->index_offset) + sizeof(struct bom_index_header) + sizeof(struct bom_index) * ntohl(index_header->count);
-    if (resize) {
+    size_t new_index_length = sizeof(struct bom_index_header) + sizeof(struct bom_index) * (ntohl(index_header->count) + 1);
+    if (new_index_length > ntohl(header->index_length) - (sizeof(struct bom_index) * 2) ) {
         ptrdiff_t index_delta = sizeof(struct bom_index);
         _bom_address_resize(context, index_point, index_delta);
+        header->index_length = htonl(ntohl(header->index_length) + sizeof(struct bom_index));        
     }
 
     /* Insert data at the very end. */
@@ -278,9 +273,6 @@ bom_index_add(struct bom_context *context, const void *data, size_t data_len)
 
     /* Update length for newly added index. */
     index_header->count = htonl(ntohl(index_header->count) + 1);
-    if (resize) {
-        header->index_length = htonl(ntohl(header->index_length) + sizeof(struct bom_index));
-    }
     header->block_count++;
 
     return added_index;
--- a/Libraries/libbom/Sources/bom.c
+++ b/Libraries/libbom/Sources/bom.c
@@ -21,7 +21,6 @@
 struct bom_context {
     struct bom_context_memory memory;
     unsigned int iteration_count;
-    unsigned int preallocated_index_count;
 };
 
 struct bom_context *
@@ -34,7 +33,6 @@ _bom_alloc(struct bom_context_memory memory)
 
     context->memory = memory;
     context->iteration_count = 0;
-    context->preallocated_index_count = 0;
 
     return context;
 }
@@ -47,7 +45,6 @@ bom_alloc_size(struct bom_context_memory memory, uint32_t index_count)
         return NULL;
     }
 
-    context->preallocated_index_count = index_count;
 
     size_t header_size = sizeof(struct bom_header);
     size_t index_size = sizeof(struct bom_index_header);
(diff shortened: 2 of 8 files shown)


Libraries/libbom/Sources/bom.c
```diff
@@ -33,21 +34,30 @@ _bom_alloc(struct bom_context_memory memory)
...
     struct bom_context *context = _bom_alloc(memory);
     if (context == NULL) {
         return NULL;
     }
 
+    context->preallocated_index_count = index_count;
+
     size_t header_size = sizeof(struct bom_header);
     size_t index_size = sizeof(struct bom_index_header);
-    size_t freelist_size = sizeof(struct bom_index_header) + sizeof(struct bom_index) * 2;
```
grp: IIRC, the value here wasn't arbitrary — I think I saw that some tools didn't handle files with a free list of size less than `2`. So it might be worth using `(2 + index_count)` here and pass `0` as the default, so you can't accidentally make an invalid file.

Libraries/libbom/Sources/bom.c
```diff
@@ -245,9 +255,16 @@ bom_index_add(struct bom_context *context, const void *data, size_t data_len)
     struct bom_index_header *index_header = (struct bom_index_header *)((void *)header + ntohl(header->index_offset));
 
     /* Insert index at the end of the list. */
+    bool resize = true;
+    if (context->preallocated_index_count > 0) {
```
grp: Why store this on the context rather than checking if the free list is empty? That let this optimization take advantage of a free list in a loaded BOM.

Libraries/libbom/Headers/bom/bom.h
```diff
@@ -42,6 +42,9 @@ struct bom_context *
 bom_alloc_empty(struct bom_context_memory memory);
 
 struct bom_context *
+bom_alloc_size(struct bom_context_memory memory, uint32_t index_count);
```
grp: I'd just merge this extra parameter into `bom_alloc_empty()`, easy enough to just pass zero if you don't know the size.

Libraries/libbom/Headers/bom/bom.h
```diff
@@ -90,6 +93,9 @@ struct bom_tree_context *
 bom_tree_alloc_empty(struct bom_context *context, const char *variable_name);
 
 struct bom_tree_context *
+bom_tree_alloc_size(struct bom_context *context, const char *variable_name, uint32_t index_count);
```
grp: I'm a little worried about this one. The semantics are different than from the BOM as a whole: while for a BOM the `index_count` is a hint, it's a hard requirement for a BOM tree. Since the BOM tree doesn't store a free list, the tree is malformed with empty entries. That isn't obvious from the API.

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -60,6 +67,17 @@ class Writer {
...
+    /*
+     * Add a rendition for a facet, optimized for fast editing of CAR files
+     */
+    void addRendition(void *key, size_t key_len, void *value, size_t value_len);
+
+    /*
+     * The key format, optional and determined automatically if omitted.
+     */
+    ext::optional<struct car_key_format *> &keyfmt()
+    { return _keyfmt; }
```
grp: I don't see these used anywhere in this PR — are you using this to copy information from another car file?

The current architecture of separate `Reader` and `Writer` objects is probably not the best one if you're trying to make small modifications to existing files, but my understanding was that wasn't being done or likely needed. If that's going to be needed, it might be worth reconsidering the architecture of `libcar` so the internals aren't so widely exposed.

grp:
Also, this PR has the best branch name. 🚗 💨

Libraries/libcar/Sources/Writer.cpp
```diff
@@ -48,6 +48,12 @@ addRendition(Rendition const &rendition)
     }
 }
 
+void Writer::
+addRendition(void *key, size_t key_len, void *value, size_t value_len)
+{
+  _raw_renditions.push_back({key, key_len, value, value_len});
```
kastiglione: `emplace_back`?

marcsalem: I guess emplace_back is faster, so sticking with the theme, I'll change it.

Libraries/libbom/Sources/bom.c
marcsalem: Sure, I could calculate out the new size of the index list and compare it to index_length in the bom header. But that is a lot more work compared to keeping track in a counter.

Also, the size calculation would have to compensate for those two extra slots that are needed empty at the end, which means keeping track of this special requirement in more than one place.

Libraries/libbom/Sources/bom.c
marcsalem: ok

Libraries/libbom/Headers/bom/bom.h
marcsalem: ok

Libraries/libbom/Sources/bom.c
grp: True on both counts. But the advantage is that you're taking advantage of the format itself, which makes this feature less a specific optimization for the actool use case, and more filling out the general libbom feature set.

Ktwu: I think this is fine for now; optimizations can happen in other PRs.

Libraries/libbom/Headers/bom/bom.h
marcsalem: This one is a little hard to follow, but in this case the count is also just a shortcut.

The allocated size of the "path indexes" is still recorded as the size of the bom_index saved to tree->child. This is recovered every time the tree is added to via:

   `struct bom_tree_entry *paths = (struct bom_tree_entry *)bom_index_get(tree_context->context, paths_index, &paths_length);`

Libraries/libbom/Sources/bom.c
```diff
@@ -33,21 +34,30 @@ _bom_alloc(struct bom_context_memory memory)
...
     context->iteration_count = 0;
+    context->preallocated_index_count = 0;
 
     return context;
 }
 
 struct bom_context *
 bom_alloc_empty(struct bom_context_memory memory)
 {
+  return bom_alloc_size(memory, 2);
```
Ktwu: Nit:  indention is wrong.

Libraries/libbom/Sources/bom.c
```diff
@@ -268,7 +285,9 @@ bom_index_add(struct bom_context *context, const void *data, size_t data_len)
 
     /* Update length for newly added index. */
     index_header->count = htonl(ntohl(index_header->count) + 1);
-    header->index_length = htonl(ntohl(header->index_length) + sizeof(struct bom_index));
+    if (resize) {
+      header->index_length = htonl(ntohl(header->index_length) + sizeof(struct bom_index));
```
Ktwu: Why is this only needed for when a resize happens?

Libraries/libbom/Sources/bom_tree.c
```diff
@@ -202,19 +213,28 @@ bom_tree_add(struct bom_tree_context *tree_context, const void *key, size_t key_
     int tree_index = bom_variable_get(tree_context->context, tree_context->variable_name);
     struct bom_tree *tree = (struct bom_tree *)bom_index_get(tree_context->context, tree_index, NULL);
 
-    /* Make room for the new index, extending the size as necessary. */
     int paths_index = ntohl(tree->child);
-    bom_index_append(tree_context->context, paths_index, sizeof(struct bom_tree_entry_indexes));
+    bool resize;
+    if (tree_context->preallocated_index_count > 0) {
+      tree_context->preallocated_index_count -= 1;
```
Ktwu: Nit:  indentation is off.

Libraries/libbom/Sources/bom_tree.c
```diff
@@ -202,19 +213,28 @@ bom_tree_add(struct bom_tree_context *tree_context, const void *key, size_t key_
     int tree_index = bom_variable_get(tree_context->context, tree_context->variable_name);
     struct bom_tree *tree = (struct bom_tree *)bom_index_get(tree_context->context, tree_index, NULL);
 
-    /* Make room for the new index, extending the size as necessary. */
     int paths_index = ntohl(tree->child);
-    bom_index_append(tree_context->context, paths_index, sizeof(struct bom_tree_entry_indexes));
+    bool resize;
+    if (tree_context->preallocated_index_count > 0) {
```
Ktwu: Nit:  It might be useful to abstract this logic out to something like "tryConsumePreallocatedResource" that encapsulates interacting with the index count.

Libraries/libbom/Sources/bom.c
marcsalem: Because if we are not resizing, we are using free space within the existing allocated size, which is index_length.

If we have run out of pre-allocated slots and are resizing, index_length is increasing by `sizeof(struct bom_index)`

Libraries/libbom/Sources/bom.c
marcsalem: As per Grant's suggestion, bom_alloc_empty is getting removed.

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -35,9 +35,16 @@ class Writer {
     typedef std::unique_ptr<struct bom_context, decltype(&bom_free)> unique_ptr_bom;
 
 private:
+    typedef struct {
+        void *key; size_t key_len; void *value; size_t value_len;
+    } KeyValuePair;
+
+private:
     unique_ptr_bom _bom;
+    ext::optional<struct car_key_format *> _keyfmt;
```
Ktwu: Why not typedef car_key_format so you don't have to keep writing 'struct'?

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -60,6 +67,17 @@ class Writer {
      */
     void addRendition(Rendition const &rendition);
 
+    /*
+     * Add a rendition for a facet, optimized for fast editing of CAR files
+     */
+    void addRendition(void *key, size_t key_len, void *value, size_t value_len);
```
Ktwu: The mixture of camel case and underscores makes me a little sad :(

Libraries/libbom/Sources/bom.c
marcsalem: preallocated_index_count is the optimization, but I'll remove it anyway so that the code can modify an existing file in the future more easily.

Libraries/libcar/Headers/car/Writer.h
marcsalem: Keeping with the style of the existing code. There are no typedefs used in bom_format.h or car_format.h

Libraries/libcar/Headers/car/Writer.h
marcsalem: fixed

Libraries/libcar/Tests/test_Writer.cpp
```diff
@@ -98,3 +98,158 @@ TEST(Writer, TestWriter)
...
+    int height = 8;
+    int create_facet_count = 100;
+    int create_scales_count = 3;
+    int create_rendition_count = create_facet_count * create_scales_count;
+
+    /* Write out. */
+    /* Write with half the required number of pre-allocated indexes */
+    int preallocated_index_count = 6 + create_facet_count * 2 + create_rendition_count * 2;
+    auto writer_bom = car::Writer::unique_ptr_bom(bom_alloc_size(bom_context_memory(NULL, 0), preallocated_index_count / 2), bom_free);
+    // auto writer_bom = car::Writer::unique_ptr_bom(bom_alloc_size(bom_context_memory(NULL, 0), 0), bom_free);
```
Ktwu: Kill this line.

Libraries/libcar/Tests/test_Writer.cpp
```diff
@@ -98,3 +98,158 @@ TEST(Writer, TestWriter)
     EXPECT_EQ(rendition_count, 1);
 }
 
+
+TEST(Writer, TestWriter100)
```
Ktwu: omg tests :D

Libraries/libcar/Tests/test_Writer.cpp
marcsalem: Arh, done

Ktwu:
LGTM.  Will merge once the builds pass.

Libraries/libbom/Headers/bom/bom.h
```diff
@@ -39,7 +39,7 @@ bom_context_memory_file(const char *fn, bool writeable, size_t minimum_size);
 struct bom_context;
 
 struct bom_context *
-bom_alloc_empty(struct bom_context_memory memory);
+bom_alloc_size(struct bom_context_memory memory, uint32_t index_count);
```
grp: Maybe still call these `bom_alloc_empty` and `bom_tree_alloc_empty`, since they create empty BOMs? I think "size" doesn't communicate as well that it's creating a *new* BOM rather than loading one in.

Libraries/libcar/Headers/car/Writer.h
```diff
@@ -35,9 +35,16 @@ class Writer {
...
+    typedef struct {
+        void *key; size_t keyLength; void *value; size_t valueLength;
+    } KeyValuePair;
+
+private:
     unique_ptr_bom _bom;
+    ext::optional<struct car_key_format *> _keyfmt;
     std::unordered_map<std::string, Facet> _facets;
     std::unordered_multimap<uint16_t, Rendition> _renditions;
+    std::vector<KeyValuePair> _raw_renditions;
```
grp: nit: `_rawRenditions` would match the camel case

Libraries/libcar/Headers/car/Writer.h
grp: I still think this isn't quite right. If you specify a key format here, it has to manually account for all attributes in non-raw renditions! If you don't, then they will silently be written out wrong.

Are you sure that computing a key format automatically is actually slow? Even for thousands of renditions, it shouldn't be much more expensive than just iterating over the renditions, since there are only a few possible attribute identifiers.

Libraries/libcar/Tests/test_Writer.cpp
grp: Are these functional tests or performance tests?

grp:
Can this be split between the libbom and libcar changes? The former look good to me as-is.

I think e59f1154ea9adba7a111138c66c3163e7a995679 could be a good option for optimizing the car operations while still abstracting the format as libcar tries to do. I'd be interested in feedback if that approach could work for your needs before landing this — I can finish fleshing that out if you think it would.

Libraries/libcar/Headers/car/Writer.h
marcsalem: ok

Libraries/libcar/Headers/car/Writer.h
marcsalem: That's true, but if you're specifying an optional value, then you probably have a reason.

The best way to handle this I think is to catch this error in the serialize functions.

Libraries/libcar/Tests/test_Writer.cpp
marcsalem: Both. The first test checks a write with both code paths (resize and not).

The second test had one extra change which I didn't end up pushing, but I might later, that called an added API in libbom that checked that the "free list" was exactly consumed and tracked the number of resize/memmoves ops via counters in the context.

Libraries/libbom/Headers/bom/bom.h
marcsalem: Honestly, I initially thought it was better to leave bom_alloc_empty() and add a new function so that we don't churn the existing API.

I'm going to go back to my initial instinct.

marcsalem:
e59f1154ea9adba7a111138c66c3163e7a995679 looks fine, maybe this can happen as a pull request after this one lands?

I think the more interesting change would be to modify the car::Writer class to be able to create/alloc the bom itself with the correct number of indexes preallocated, or maybe to add an API to libbom to be able to resize the free list by a specified amount.

That way actool can benefit from these speed-ups.

PR #111, opened 2016-08-30
Add basic PlistBuddy

onhachoe:


Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+    auto loc = keyPathString.find(':');
+    if (loc == std::string::npos) {
+        if (keyPathString.size() > 0) {
+            keyPath.push(keyPathString);
+        }
+    } else {
+        if (loc > 0) {
+            keyPath.push(keyPathString.substr(0, loc));
+        }
+        parseCommandKeyPathString(keyPathString.substr(loc + 1), keyPath);
```
kastiglione: Does `istringstream` have downsides?

```cpp
std::istringstream source(keyPathString);
for (std::string key; std::getline(source, key, ':'); ) {
  keyPath.push(key);
}
```

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+    if (tokens.size() < 1) {
+        return true;
+    }
+
+    std::string command = tokens[0];
+    if (command == "Print") {
+        std::queue<std::string> keyPath;
+        if (tokens.size() < 2) {
+            parseCommandKeyPathString("", keyPath);
```
kastiglione: From the looks of `parseCommandKeyPathString`, this is a no-op.

Libraries/plist/CMakeLists.txt
```diff
@@ -69,6 +69,10 @@ add_executable(plutil Tools/plutil.cpp)
 target_link_libraries(plutil plist util)
 install(TARGETS plutil DESTINATION usr/bin)
 
+add_executable(PlistBuddy Tools/PlistBuddy.cpp)
+target_link_libraries(PlistBuddy plist util edit)
```
grp: I don't think libedit is available on Linux, and I'd assume you'd want to talk to lawyers before touching readline, so maybe add a submodule for https://github.com/antirez/linenoise or https://github.com/troglobit/editline?

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+        return libutil::Options::NextString(&_command, args, it);
+    } else if (!arg.empty() && arg[0] != '-') {
+        _input = arg;
+        return std::make_pair(true, std::string());
+    } else {
+        return std::make_pair(false, "unknown argument " + arg);
+    }
+}
+
+plist::ObjectType parseType(std::string &typeString) {
```
grp: nit: `static`, newline after return type, `std::string const &` for type

Libraries/plist/Tools/PlistBuddy.cpp
grp: Streams are an unfortunate part of the C++ stdlib, full of operator overloading and mutable references usually avoided in xcbuild. I also avoided `std::getline()` specifically for both a misleadingly name and mutable references. It would be nice if there was a `std::split()` of some kind.

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+    fprintf(stderr, "\noptions:\n");
+    fprintf(stderr, INDENT "-c \"<command>\" command to execute, otherwise run in interactive mode\n");
+    fprintf(stderr, INDENT "-x output will be in xml plist format\n");
+    fprintf(stderr, INDENT "-h print help including commands\n");
+#undef INDENT
+
+    return (error.empty() ? 0 : -1);
+}
+
+static void parseCommandKeyPathString(const std::string &keyPathString, std::queue<std::string> &keyPath) {
```
grp: nit: newline after return type, capitalize function name, `std::string const &`, and `std::queue<std::string> *` for mutating

Libraries/plist/Tools/PlistBuddy.cpp
grp: Also, is it guaranteed that the tail recursion here is optimized? If not, might be worth making the iteration explicit.

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+{
+}
+
+Options::
+~Options()
+{
+}
+
+std::pair<bool, std::string> Options::
+parseArgument(std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it)
```
grp: (`libutil::Options` was changed to use `ext::optional` recently, so this needs to be updated; see `plutil.cpp` for an example.)

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+            array->remove(index);
+            return true;
+        }
+        default:
+            std::cerr << "Invalid key path (removing value on non-collection object)" << std::endl;
+            return false;
+    }
+}
+
+static void CommandHelp()
```
grp: nit: newline after type, maybe move this near `Help()`?

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+static const char *prompt(EditLine *e) {
+  return "Command: ";
+}
+
+int
+main(int argc, char **argv)
+{
+    std::vector<std::string> args = std::vector<std::string>(argv + 1, argv + argc);
+    auto filesystem = std::unique_ptr<libutil::Filesystem>(new libutil::DefaultFilesystem());
```
grp: nit: this can be just `libutil::DefaultFilesystem filesystem = libutil::DefaultFilesystem();` and skip the `std::unique_ptr`; see 11725f9b3a541ac97f568b679ad07719981f68b8

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+        while (keepReading) {
+            /* libedit calls will go here */
+            line = el_gets(el, &count);
+            if (count > 0) {
+              history(cmdhistory, &ev, H_ENTER, line);
+
+              /* strip trailing newline */
+              std::string strline = std::string(line, 0, count - 1);
+              keepReading = ProcessCommand(filesystem.get(), options.input(), options.xml(), options.input(), root, strline);
```
grp: nit: indentation here is off

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+            std::cerr << "Add command requires KeyPath" << std::endl;
+            return true;
+        }
+        std::queue<std::string> keyPath;
+        parseCommandKeyPathString(tokens[1], keyPath);
+        Delete(root.object.get(), keyPath);
+    } else if (command == "Help") {
+        CommandHelp();
+    } else {
+        std::cerr << "Unrecognized command" << std::endl;
```
grp: nit: use `fprintf` instead of C++ streams

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,572 @@
...
+    }
+    return true;
+}
+
+static const char *prompt(EditLine *e) {
```
grp: nit: `static char const *`, newline after return type, indentation is off

Libraries/plist/Tools/PlistBuddy.cpp
onhachoe: I sure too hope there was std::split of some kind with vanilla cpp (boost should have one, but that makes things too complicated i think), but there's none to my knowledge and best bet is to use the stringstream, but it has problem of persisting empty string at the end of iteration.

I personally think this is good as is, as making it iterative would just bloat up the logic.

grp: One advantage of iteration with `find()` (and without `std::string_view`) is that it's O(N), while this is O(N^2) due to the `substr()`.

Libraries/plist/CMakeLists.txt
```diff
@@ -69,6 +69,11 @@ add_executable(plutil Tools/plutil.cpp)
 target_link_libraries(plutil plist process util)
 install(TARGETS plutil DESTINATION usr/bin)
 
+add_executable(PlistBuddy Tools/PlistBuddy.cpp ${CMAKE_SOURCE_DIR}/ThirdParty/linenoise/linenoise.c)
+target_include_directories(PlistBuddy PRIVATE "${CMAKE_SOURCE_DIR}/ThirdParty/linenoise")
+target_link_libraries(PlistBuddy plist util)
+install(TARGETS PlistBuddy DESTINATION usr/bin)
```
grp: This is the first thing that requires `git submodule init` to build — @strager, is that still an issue?

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+#include <sstream>
+#include <string>
+#include <queue>
+
+typedef struct {
+    std::unique_ptr<plist::Object> object;
+} RootObjectContainer;
+
+class Options {
+public:
```
grp: nit: remove this

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+        }
+    } else {
+        if (loc > 0) {
+            keyPath->push(keyPathString.substr(0, loc));
+        }
+        ParseCommandKeyPathString(keyPathString.substr(loc + 1), keyPath);
+    }
+}
+
+static std::string parseCommandValueString(std::vector<std::string>::const_iterator begin, std::vector<std::string>::const_iterator end) {
```
grp: nit: newline for function name, capitalize name of static function

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+    for (auto it = begin; it != end; it++) {
+        if (it != begin) {
+            sstream << ' ';
+        }
+        sstream << it->c_str();
+    }
+
+    std::string valueString = sstream.str();
+    if (valueString.front() == '\"' && valueString.back() == '\"') {
+        // strip double quote if necessary
```
grp: nit: `/* */`

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+        valueString = valueString.substr(1, valueString.size() - 2);
+    }
+    return valueString;
+}
+
+static plist::Object *getObjectAtKeyPath(plist::Object *object, std::queue<std::string> &remainingKeys, bool leaveLastKey = true) {
```
grp: nit: newline + capitalize function name

grp: nit: pass `remainingKeys` as a pointer for mutation

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+        fprintf(stderr, "Invalid key path (indexing into null object)\n");
+        return nullptr;
+    }
+
+    std::string currentKey = remainingKeys.front();
+
+    plist::Object *subObject = nullptr;
+    switch (object->type()) {
+        case plist::ObjectType::Dictionary:
+            subObject = plist::CastTo<plist::Dictionary>(object)->value(currentKey);
```
grp: Rather than the `switch`, you could do this:

    if (plist::Dictionary *dictionary = plist::CastTo<plist::Dictionary>(object)) {
        subObject = dictionary->value(currentKey);
    } else if (plist::Array *array = plist::CastTo<plist::Array>(object)) {
        /* ... */
    } else {
        /* ... */
    }

But not too different either way...

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+    if (!subObject) {
+        return object;
+    } else {
+        remainingKeys.pop();
+        return getObjectAtKeyPath(subObject, remainingKeys, leaveLastKey);
+    }
+}
+
+static bool
+Print(plist::Object *object, std::queue<std::string> &keyPath, bool xml, libutil::Filesystem *fileSystem = nullptr, std::string path = "-")
```
grp: nit: pass `keyPath` as a pointer if it's going to be mutated.
nit: `fileSystem` -> `filesystem`

When would the filesystem not be available?

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+            array->insert(index, std::move(newObj));
+            return true;
+        }
+        default:
+            fprintf(stderr, "Invalid key path (setting value on non-collection object)\n");
+            return false;
+    }
+}
+
+static bool Clear(RootObjectContainer &root, plist::ObjectType clearType) {
```
grp: nit: pass a pointer for mutation

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+            root.object = plist::Data::New();
+            break;
+        default:
+            fprintf(stderr, "Unsupported type\n");
+            return false;
+    }
+    return true;
+}
+
+static bool Delete(plist::Object *object, std::queue<std::string> &keyPath) {
```
grp: nit: pointer to mutate `keyPath`

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+    Options options;
+    std::pair<bool, std::string> result = libutil::Options::Parse<Options>(&options, args);
+    if (!result.first) {
+        return Help(result.second);
+    }
+
+    if (options.help() || options.input().empty()) {
+        Help();
+        return 1;
```
grp: Maybe `return Help("no input provided");`?

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+        ProcessCommand(&filesystem, options.input(), options.xml(), options.input(), root, *options.command());
+        // saves
+        std::queue<std::string> keyPath;
+        Print(root.object.get(), keyPath, options.xml(), &filesystem, options.input());
+    } else {
+        const char *line;
+        bool keepReading = true;
+
+        while (keepReading) {
+            /* linenoise calls will go here */
```
grp: Already done?

Libraries/plist/Tools/PlistBuddy.cpp
onhachoe: Not a big fan of assigning stuff within the if clause, but can change if you insist.
But switch-case saves CPU :P

Libraries/plist/CMakeLists.txt
onhachoe: hmm, although this is convenient CLI feature, if this would cause problem, we can go without the library.
Just kept it in place since @jasonprado had put it in place before.

Libraries/plist/Tools/PlistBuddy.cpp
onhachoe: ProcessCommand->Print will call this without filesystem, it can be changed to call with this argument passed in if you want, but don't really see the point in changing this.

Libraries/plist/Tools/PlistBuddy.cpp
onhachoe: Well, this path is taken also when user types in "Help"
In either case, it is kind of self evident, and not sure if breaking this into two separate if's is worth.

Libraries/plist/CMakeLists.txt
grp: Would it work to make it optional? Check if `${CMAKE_SOURCE_DIR}/ThirdParty/linenoise` exists, and compile it in if it does. Can do the same check in the code with `__has_include(<linenoise.h>)` (or pass it in as a `#define` from CMake).

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+static bool
+Delete(plist::Object *object, std::queue<std::string> *keyPath) {
+    plist::Object *target = GetObjectAtKeyPath(object, keyPath, true);
+    if (!target || keyPath->size() != 1) {
+        fprintf(stderr, "Invalid key path (target object not found)\n");
+        return false;
+    }
+
+    std::string targetKey = keyPath->front();
```
grp: nit: This could be `std::string const &`.

Libraries/plist/CMakeLists.txt
strager: Make it optional if it's easy to do. If not, Facebook can deal with the required dependency.

grp: Made optional in 96ed319dcc01244e73dd4dd872b9392c1df1c6ef. Falls back to `fgets()` when not available.

PR #112, opened 2016-08-31
Clean up code for build settings.

grp:
Settings was some of the oldest code, so needed some work.

 - Modernize outdated syntax.
 - Rewrite configuration file parsing.
 - Support additional failure cases.


Libraries/pbxsetting/Headers/pbxsetting/Setting.h
```diff
@@ -14,6 +14,7 @@
 #include <pbxsetting/Value.h>
```
Ktwu: LGTM

Libraries/pbxsetting/Headers/pbxsetting/Value.h
```diff
@@ -13,6 +13,7 @@
 #include <memory>
```
Ktwu: Yay cleanup.  +2.

Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -269,8 +304,8 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
             projectActionEnvironment.insertFront(level, false);
         }
 
-        projectConfigurationFile = LoadConfigurationFile(projectConfiguration, projectActionEnvironment);
-        if (projectConfigurationFile != nullptr) {
+        projectConfigurationFile = LoadConfigurationFile(Filesystem::GetDefaultUNSAFE(), projectActionEnvironment, workingDirectory, projectConfiguration);
```
Ktwu: GetDefaultUNSAFE()?  That's a fun function; why's it unsafe?

Ktwu: Since you're using this further down, too, maybe cache the Unsafe filesystem in a local variable?

Libraries/pbxsetting/Sources/XC/Config.cpp
```diff
@@ -8,40 +8,205 @@
...
+Config(std::string const &path, std::vector<Entry> const &contents) :
+    _path    (path),
+    _contents(contents)
 {
-    std::pair<bool, Level> result = ConfigFile().open(path, environment, error);
-    if (!result.first) {
-        return nullptr;
+}
+
+Config::
```
Ktwu: Nit:  Get rid of empty destructor?

Libraries/pbxbuild/Sources/Target/Environment.cpp
grp: From 8e36066203c13d0dee7911c8e8bc168eb8d1da67. It's unsafe because calling code can't pass down the filesystem — it always uses a `DefaultFilesystem` (i.e. the real one), even from a unit test using a `MemoryFilesystem`. It can even modify the filesystem without declaring a (non-const) `Filesystem *` parameter in its API.

The issue in this case is that of the main parts of the build:

 1. Load developer directory, platforms, SDKs, toolchains.
 2. Load the workspace and project.
 3. Determine build settings, compute invocations & auxiliary files.
 4. Generate Ninja or run the invocations' commands.

Parts 1, 2, and 4 clearly need filesystem access, but part 3 only occasionally does. My hope is that the filesystem can be removed from part 3 completely, to isolate and test the complex logic in computing settings & invocations. The plan for that is split on reads and writes:

 - Reads get pre-loaded in phase 2, as part of loading the project structure. This is complicated for xcconfigs specifically because build settings (including transient build-specific ones, like command-line build settings) can affect the path the xcconfig is loaded from, and phase 2 is build-independent.
 - Writes get deferred from phase 3 to phase 4, like how auxiliary files are already described in invocations rather than written out directly by the resolving code.

Ktwu: Ok, this makes a lot of sense.  Thanks for the explanation.

Libraries/pbxbuild/Headers/pbxbuild/Target/Environment.h
```diff
@@ -50,15 +50,30 @@ class Environment {
     std::vector<std::string>                 _architectures;
```
Ktwu: I know little about pbxsettings, so a lot of this is over my head.  That said, tests are good, and no obvious bugs stood out to me, so LGTM.

PR #113, opened 2016-09-01
WIP Generate VFS for importing header maps.

grp:
The main issue here is that the generated VFS is not per-target! It's
an auxiliary file that contains information from all targets and goes
in a target-independent directory. This doesn't match with either the
auxiliary file model or more generally the entire build structure:

 - Can anything specific to a build affect the VFS? If so, what if
   separarate builds are running in parallel from the same project?
 - If two targets in a build are from the same project, they should
   only need to generate the VFS once, and not more than once.


sas:
Does this still need to stay open or can we abandon it?

vhbit:
This might be relevant for #237

PR #114, opened 2016-09-03
Preload external build configuration files at project load time.

grp:
This avoids requiring filesystem access to compute the target environment,
and fixes the layering of build configurations when determining the target
SDK for the build.

### DIFF
--- a/Libraries/pbxbuild/Headers/pbxbuild/Target/Environment.h
+++ b/Libraries/pbxbuild/Headers/pbxbuild/Target/Environment.h
@@ -50,10 +50,6 @@ private:
     std::vector<std::string>                 _architectures;
 
 private:
-    ext::optional<pbxsetting::XC::Config>    _projectConfigurationFile;
-    ext::optional<pbxsetting::XC::Config>    _targetConfigurationFile;
-
-private:
     std::string                              _workingDirectory;
     std::unordered_map<pbxproj::PBX::BuildFile::shared_ptr, std::string> _buildFileDisambiguation;
 
@@ -70,8 +66,6 @@ private:
         std::shared_ptr<pbxsetting::Environment> const &environment,
         std::vector<std::string> const &variants,
         std::vector<std::string> const &architectures,
-        ext::optional<pbxsetting::XC::Config> const &projectConfigurationFile,
-        ext::optional<pbxsetting::XC::Config> const &targetConfigurationFile,
         std::string const &workingDirectory,
         std::unordered_map<pbxproj::PBX::BuildFile::shared_ptr, std::string> const &buildFileDisambiguation);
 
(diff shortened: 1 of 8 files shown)


Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -282,59 +276,67 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     pbxproj::XC::BuildConfiguration::shared_ptr targetConfiguration;
     ext::optional<pbxsetting::XC::Config> projectConfigurationFile;
     ext::optional<pbxsetting::XC::Config> targetConfigurationFile;
+
     {
-        // FIXME(grp): $(SRCROOT) must be set in order to find the xcconfig, but we need the xcconfig to know $(SDKROOT). So this can't
-        // use the default level order, because $(SRCROOT) comes below $(SDKROOT). Hack around this for now with a synthetic environment.
-        // It's also in the wrong order because project settings should be below the SDK, but are needed to *load* the xcconfig.
+        /*
+         * Create a synthetic build setting environment to determine the SDK to use. The real build
```
Ktwu: So...using a synthetic environment is no longer a hack?

grp: Yeah. I found out a while ago that it was expected behavior, but I forgot to update this comment.

Libraries/pbxbuild/Sources/WorkspaceContext.cpp
```diff
@@ -143,7 +152,38 @@ LoadWorkspaceProjects(Filesystem const *filesystem, std::vector<pbxproj::PBX::Pr
...
 
 static void
-LoadNestedProjects(Filesystem const *filesystem, std::vector<pbxproj::PBX::Project::shared_ptr> *projects, pbxsetting::Environment const &baseEnvironment, std::vector<pbxproj::PBX::Project::shared_ptr> const &rootProjects)
+LoadConfigurationFiles(
+    Filesystem const *filesystem,
+    std::unordered_map<pbxproj::XC::BuildConfiguration::shared_ptr, pbxsetting::XC::Config> *configs,
+    pbxsetting::Environment const &environment,
+    pbxproj::XC::ConfigurationList::shared_ptr const &configurationList)
+{
+    if (configurationList == nullptr) {
```
Ktwu: Isn't configurationList a reference?  I didn't think it was reasonable to check whether such types were null.

Libraries/pbxbuild/Headers/pbxbuild/Target/Environment.h
```diff
@@ -50,10 +50,6 @@ class Environment {
     std::vector<std::string>                 _architectures;
```
Ktwu: Since it's been a while, LGTM so I don't block you.

Ktwu: Once Travis succeeds, go ahead and merge.

Libraries/pbxbuild/Sources/WorkspaceContext.cpp
grp: It's a reference to a `std::shared_ptr`, for some reason xcbuild has a lot of `typedef`s for that rather than using template syntax. The pointer can be `nullptr` even if the reference to it can't be.

Libraries/pbxbuild/Headers/pbxbuild/Target/Environment.h
grp: Travis is failing from libpng install, @marcsalem could you take a look?

marcsalem: Sure, the mirror I used seems to remove old versions as soon as there is an
update

PR #116, opened 2016-09-09
[xcsdk] Allow xcrun to work with only --toolchains defined when finding tools

Ktwu:
If the tool isn't given an SDK, then it doesn't make sense to fail the
tool search if the asked-for tool lives inside one of the given
toolchains.

This also changes how 'xcrun' operates -- if there's no SDK target that can be found,
then that's OK so long as there's some sort of toolchain path that can be searched.

Suggestions for how to test this?

Change-Id: I17f8cd3fbe7d02ad31a6cdbbc016f5ab6ffe2f3c

### DIFF
--- a/Libraries/pbxbuild/Sources/Target/Environment.cpp
+++ b/Libraries/pbxbuild/Sources/Target/Environment.cpp
@@ -440,7 +440,7 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     }
 
     /* Tool search directories. Use the toolchains just discovered. */
-    std::vector<std::string> executablePaths = sdk->executablePaths(toolchains);
+    std::vector<std::string> executablePaths = buildEnvironment.sdkManager()->executablePaths(sdk, toolchains);
 
     auto buildRules = Target::BuildRules::Create(buildEnvironment.specManager(), specDomains, target);
     auto buildFileDisambiguation = BuildFileDisambiguation(target);
--- a/Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
+++ b/Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
@@ -80,7 +80,7 @@ public:
     /*
      * Default settings for the contents of the developer root.
      */
-    pbxsetting::Level computedSettings(void) const;
+    pbxsetting::Level computedSettings() const;
 
 public:
     /*
@@ -88,6 +88,12 @@ public:
      * root, rather than within a toolchain or a platform.
      */
     std::vector<std::string> executablePaths() const;
+
+    /*
+     * Conglomeration of executable paths that optionally includes extra toolchains
+     * and the paths from an SDK target.
+     */
+    std::vector<std::string> executablePaths(xcsdk::SDK::Target::shared_ptr const &target, std::vector<xcsdk::SDK::Toolchain::shared_ptr> const &toolchains) const;
 
 public:
     /*
--- a/Libraries/xcsdk/Sources/SDK/Manager.cpp
+++ b/Libraries/xcsdk/Sources/SDK/Manager.cpp
@@ -123,6 +123,27 @@ executablePaths() const
     };
 }
 
+std::vector<std::string> Manager::
+executablePaths(xcsdk::SDK::Target::shared_ptr const &target, std::vector<xcsdk::SDK::Toolchain::shared_ptr> const &toolchains) const
+{
+    std::vector<std::string> paths;
+
+    if (target != nullptr) {
+        std::vector<std::string> targetPaths = target->executablePaths(toolchains);
+        paths.insert(paths.end(), targetPaths.begin(), targetPaths.end());
+    } else {
+        for (Toolchain::shared_ptr const &toolchain : toolchains) {
+            std::vector<std::string> toolchainPaths = toolchain->executablePaths();
+            paths.insert(paths.end(), toolchainPaths.begin(), toolchainPaths.end());
+        }
+    }
+
+    std::vector<std::string> managerPaths = this->executablePaths();
+    paths.insert(paths.end(), managerPaths.begin(), managerPaths.end());
+
+    return paths;
+}
+
 std::shared_ptr<Manager> Manager::
 Open(Filesystem const *filesystem, std::string const &path, ext::optional<Configuration> const &configuration)
 {
--- a/Libraries/xcsdk/Tools/xcrun.cpp
+++ b/Libraries/xcsdk/Tools/xcrun.cpp
@@ -295,94 +295,118 @@ main(int argc, char **argv)
         fprintf(stderr, "verbose: using developer root '%s'\n", manager->path().c_str());
     }
 
+    bool showSDKValue = options.showSDKPath() ||
+        options.showSDKVersion() ||
+        options.showSDKBuildVersion() ||
+        options.showSDKPlatformPath() ||
+        options.showSDKPlatformVersion();
+
     /*
      * Determine the SDK to use.
      */
-    xcsdk::SDK::Target::shared_ptr target = manager->findTarget(SDK.value_or("macosx"));
-    if (target == nullptr) {
-        if (SDK) {
-            fprintf(stderr, "error: unable to find sdk '%s'\n", SDK->c_str());
-        } else {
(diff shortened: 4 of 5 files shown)


Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -440,7 +440,10 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     }
 
     /* Tool search directories. Use the toolchains just discovered. */
-    std::vector<std::string> executablePaths = sdk->executablePaths(toolchains);
+    std::vector<std::string> executablePaths;
+    if (auto manager = sdk->manager()) {
```
grp: Could you use `buildEnvironment.sdkManager()` here instead? It's always available and doesn't depend on the unfortunate nullable backpointers in the SDK objects.

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
```diff
@@ -89,6 +89,12 @@ class Manager {
      */
     std::vector<std::string> executablePaths() const;
 
+    /*
+     * Conglomeration of executable paths that optionally includes extra toolchains
+     * and the paths from an SDK target.
+     */
+    std::vector<std::string> executablePaths(std::vector<xcsdk::SDK::Toolchain::shared_ptr> &toolchains, xcsdk::SDK::Target::shared_ptr &target) const;
```
grp: nit: `const &` for both parameters

Libraries/xcsdk/Sources/SDK/Manager.cpp
```diff
@@ -123,6 +123,27 @@ executablePaths() const
...
+{
+    std::vector<std::string> paths;
+
+    for (Toolchain::shared_ptr const &toolchain : toolchains) {
+        std::vector<std::string> toolchainPaths = toolchain->executablePaths();
+        paths.insert(paths.end(), toolchainPaths.begin(), toolchainPaths.end());
+    }
+
+    if (target != nullptr) {
+        std::vector<std::string> targetPaths = target->executablePaths();
```
grp: I think this is changing behavior: before, if you used `overrideToolchains`, those paths would be used *instead* of the target's toolchains. Now, the passed-in toolchains are used *in addition* to the target's toolchains, since the target's toolchains are included in its paths. Even with that, I think the ordering is backwards: the passed-in override toolchains are now ending up *before* the target's default toolchains.

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -298,17 +298,70 @@ main(int argc, char **argv)
     /*
      * Determine the SDK to use.
      */
-    xcsdk::SDK::Target::shared_ptr target = manager->findTarget(SDK.value_or("macosx"));
-    if (target == nullptr) {
-        if (SDK) {
-            fprintf(stderr, "error: unable to find sdk '%s'\n", SDK->c_str());
+    const std::string defaultSDK = "macosx";
+    xcsdk::SDK::Target::shared_ptr target = manager->findTarget(SDK.value_or(defaultSDK));
```
grp: I think it should be OK to not specify any `-sdk` and have no default, but if you do specify an `-sdk`, not finding that SDK should be an error. Sort of like this:

    xcsdk::SDK::Target::shared_ptr target = nullptr;
    if (SDK) {
        target = manager->findTarget(*SDK);
        if (target == nullptr) {
             /* error */
        }
    } else {
        target = target = manager->findTarget(defaultSDK);
        /* nullptr target is not an error */
    }

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -298,17 +298,70 @@ main(int argc, char **argv)
...
+        } else if (options.showSDKPlatformVersion()) {
+            if (auto platform = target->platform()) {
+                printf("%s\n", platform->version().c_str());
+            } else {
+                fprintf(stderr, "error: sdk has no platform\n");
+                return -1;
+            }
+        }
+
+        return 0;
```
grp: The early return here makes the flow a little non-obvious to me, maybe wrap the code below in an `else` block?

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
grp: I'm not sure if this structure is better for this method than before: I think in most cases it's an error for `target` to be `nullptr` here.

The fact that it's valid in `xcrun` seems like part of the command-line API of that tool, rather than something worth baking into `xcsdk`. With the old method as-is, you'd have something like this:

    if (target != nullptr) {
        executablePaths.extend(target.executablePaths(toolchains));
    } else {
        for (xcsdk::SDK::Toolchain::shared_ptr const &toolchain : toolchains) {
            executablePaths.extend(toolchain.executablePaths());
    }

I don't think that's so bad?

Libraries/pbxbuild/Sources/Target/Environment.cpp
Ktwu: Sure.

Libraries/xcsdk/Tools/xcrun.cpp
Ktwu: Makes sense.

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
Ktwu: I can certainly get rid of this -- the main thing I wanted to make sure was added was all of the other stuff that target.executablePaths() adds, like the platform's paths and the manager's paths.

Libraries/xcsdk/Sources/SDK/Manager.cpp
Ktwu: Ah, thanks for catching.  I figured the given structure was really weird before -- when grabbing the target's executable paths, it seemed weird that you could override the toolchains that the SDK target used (meaning that the paths returned weren't really associated with the SDK target).  But, if that's intended, I'll rethink.

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -295,94 +295,118 @@ main(int argc, char **argv)
...
+    if (SDK) {
+        target = manager->findTarget(*SDK);
+        if (target == nullptr) {
+             printf("error: unable to find sdk: '%s'\n", SDK->c_str());
+             return -1;
+        }
+    } else {
+        target = target = manager->findTarget(defaultSDK);
+        /* nullptr target is not an error (except later on if SDK information is requested) */
+        if (showSDKValue && target == nullptr) {
```
grp: Hmm maybe move this down so the target check is closer to where it matters that target is not `nullptr`?

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -295,94 +295,118 @@ main(int argc, char **argv)
...
-        if (SDK) {
-            fprintf(stderr, "error: unable to find sdk '%s'\n", SDK->c_str());
-        } else {
-            fprintf(stderr, "error: unable to find default sdk\n");
+    const std::string defaultSDK = "macosx";
+    xcsdk::SDK::Target::shared_ptr target = nullptr;
+    if (SDK) {
+        target = manager->findTarget(*SDK);
+        if (target == nullptr) {
+             printf("error: unable to find sdk: '%s'\n", SDK->c_str());
```
grp: nit: `fprintf(stderr, `

Libraries/xcsdk/Tools/xcrun.cpp
grp: Wait, never mind, I read this backwards.

Libraries/xcsdk/Sources/SDK/Manager.cpp
grp: Yeah, it is weird... not sure of the best structure for this. You want to be able to get the target's paths with the target's toolchains, the target's paths with custom toolchains, toolchain paths with no target, and all of them with the right ordering and including the manager's paths... which leads to what you did here. So maybe that is the best after all.

grp:
Rebased & pushed.

PR #121, opened 2016-09-16
[xcbuild] Check ~/.xcsdk/xcode_dir_path when getting developer root

Ktwu:
xcrun attempts to determine the developer root so that it knows where
how to locate toolchains and sdks.
Currently the developer root is resolved through the following steps
1. Checking DEVELOPER_DIR environment variable
2. Check the /var/db/xcode_select_link
3. /usr/share/xcode-select/xcode_dir_path
(1) suffers from having to configure this env var everytime xcrun is run.
(2) and (3) suffers on needing to be root to configure on most systems.

This change adds another symlink to check for:
~/.xcsdk/xcode_dir_path

Test Plan:
1. Create symlink ~/.xcsdk/xcode_dir_path
2. xcrun -v -l --sdk test-sdk -f test-script

Change-Id: I88ed516e7264cffa263f6ed831e34083930e3337

### DIFF
--- a/Libraries/xcsdk/Sources/Environment.cpp
+++ b/Libraries/xcsdk/Sources/Environment.cpp
@@ -7,6 +7,10 @@
  of patent rights can be found in the PATENTS file in the same directory.
  */
 
+#include <pwd.h>
+#include <sys/types.h>
+#include <unistd.h>
+
 #include <xcsdk/Environment.h>
 #include <libutil/Base.h>
 #include <libutil/Filesystem.h>
@@ -16,6 +20,16 @@
 using xcsdk::Environment;
 using libutil::Filesystem;
 using libutil::FSUtil;
+
+static std::string
+UserDeveloperRootLink()
+{
+    char *home_dir = getenv("HOME");
+    if (home_dir == NULL) {
+        home_dir = getpwuid(getuid())->pw_dir;
+    }
+    return std::string(home_dir) + "/.xcsdk/xcode_select_link";
+}
 
 static std::string
 PrimaryDeveloperRootLink()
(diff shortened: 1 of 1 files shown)


Ktwu:
PR on Naris' behalf.

grp:
FWIW `xcode_select_link` is the newer name, so maybe match that for this rather than `xcode_dir_path`?

Libraries/xcsdk/Sources/Environment.cpp
```diff
@@ -17,6 +21,16 @@ using libutil::Filesystem;
 using libutil::FSUtil;
 
 static std::string
+UserDeveloperRootLink()
+{
+    char *home_dir = getenv("HOME");
+    if (home_dir == NULL) {
+        home_dir = getpwuid(getuid())->pw_dir;
+    }
+    return std::string(home_dir) + "/.xcsdk/xcode_dir_path";
```
grp: Ah, the other place in this file uses `getenv("HOME")`.

Ktwu:
When you say `xcode_select_link` is the newer name, does this mean `xcode_dir_path` is no longer supported?

grp:
It's still the fallback, but the other one is preferred if available.

Sent from my iPhone

On Sep 20, 2016, at 02:10, Ktwu <notifications@github.com> wrote:

When you say xcode_select_link is the newer name, does this mean
xcode_dir_path is no longer supported?

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/facebook/xcbuild/pull/121#issuecomment-248092128>, or mute
the thread

Ktwu:
Ok, updated the path to xcode_select_link.

grp:
How soon do you need this in? I think it might be easier to rebase this on #124 than the reverse.

Ktwu:
This doesn't need to go in soon -- my team has local changes that we're working with.

Ktwu:
Just rebased.

Libraries/xcsdk/Sources/Environment.cpp
```diff
@@ -18,6 +22,16 @@ using libutil::Filesystem;
 using libutil::FSUtil;
 
 static std::string
+UserDeveloperRootLink()
+{
+    char *home_dir = getenv("HOME");
+    if (home_dir == NULL) {
+        home_dir = getpwuid(getuid())->pw_dir;
+    }
```
grp: There's a method on `process::Context` for this now.

grp:
Landed with modifications to use `process::Context`.

grp:
Landed with modifications to use `process::Context`.

PR #123, opened 2016-09-19
Update asset compiler driver options to include and ignore new flags.

kolinkrewinkel:
There are two new arguments that need to be included in order to not fail builds:
- `--sticker-pack-identifier-prefix`
- `--product-type`

They've been added in a rudimentary way here to get around the build failures, but they're just ignored.

### DIFF
--- a/Libraries/acdriver/Headers/acdriver/Options.h
+++ b/Libraries/acdriver/Headers/acdriver/Options.h
@@ -57,6 +57,11 @@ private:
     ext::optional<std::string> _filterForDeviceModel;
     ext::optional<std::string> _filterForDeviceOsVersion;
 
+// Unimplmeneted/ignored
+private:
+    ext::optional<std::string> _stickerPackIdentifierPrefix;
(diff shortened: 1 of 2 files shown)


Libraries/acdriver/Headers/acdriver/Options.h
```diff
@@ -57,6 +57,11 @@ class Options {
     ext::optional<std::string> _filterForDeviceModel;
     ext::optional<std::string> _filterForDeviceOsVersion;
 
+// Unimplmeneted/ignored
```
grp: Rather than this comment, can add a warning to `WarnUnsupportedOptions()` in `Libraries/acdriver/Sources/CompileAction.cpp`.

grp:
One more new flag: `--sticker-pack-strings-file`.

grp:
Pushed a version of this.

kolinkrewinkel:
Awesome, thanks!

PR #124, opened 2016-09-20
Refactor process information and launching to avoid global state.

grp:
This refactors two pieces of previously global state into injectable and simulatable objects.

The first is process context, including environment variables, command-line arguments, Unix user and group, and working directory. In most cases, this state is now conceptually captured at the start of `main()`, and explicitly passed down where needed.

The second is launching subprocesses. These now only explicitly inherit the above process context, and require a process launcher to be passed down. The process launcher also has a simulated version to allow tests to stub process invocations.

This pull request should be rebased, not squashed (or merged).


Libraries/builtin/Sources/Driver.cpp
```diff
@@ -11,13 +11,13 @@
 #include <libutil/DefaultFilesystem.h>
```
Ktwu: This commit LGTM

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include <libutil/DefaultProcessContext.h>
+#include <libutil/FSUtil.h>
+
+#include <sstream>
+#include <unordered_set>
+#include <cstring>
+#include <cassert>
+
+#include <pwd.h>
+#include <grp.h>
```
Ktwu: At first blush, Grant, I thought you had your own header for some reason.

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (_NSGetExecutablePath(&buffer[0], &size) != 0) {
+        abort();
+    }
+
+    return FSUtil::NormalizePath(buffer);
+#elif defined(__linux__)
+#if __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 16
+    char const *path = reinterpret_cast<char const *>(getauxval(AT_EXECFN));
+    if (path == NULL) {
+        assert(false);
```
Ktwu: Why abort() after assert()?

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    std::string buffer;
+    buffer.resize(size);
+    if (_NSGetExecutablePath(&buffer[0], &size) != 0) {
+        abort();
+    }
+
+    return FSUtil::NormalizePath(buffer);
+#elif defined(__linux__)
+#if __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 16
```
Ktwu: nit: move this define somewhere else, like #define IS_GLIBC_2_16, since you use it more than once.

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+    char const *path = reinterpret_cast<char const *>(getauxval(AT_EXECFN));
+    if (path == NULL) {
+        assert(false);
+        abort();
+    }
+#elif defined(__GLIBC__)
+    char const *path = reinterpret_cast<char const *>(initialExecutablePath);
+#else
+#error Unsupported platform.
```
Ktwu: Maybe add "Requires GLIBC"?

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+        abort();
+    }
+
+    std::string buffer;
+    buffer.resize(size);
+    if (_NSGetExecutablePath(&buffer[0], &size) != 0) {
+        abort();
+    }
+
+    return FSUtil::NormalizePath(buffer);
```
Ktwu: If you were interested, you could rework this code so that this method used just one return statement (opposed to APPLE having its own return flow).

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+    if (char *value = getenv(variable.c_str())) {
+        return std::string(value);
+    } else {
+        return ext::nullopt;
+    }
+}
+
+std::unordered_map<std::string, std::string> DefaultProcessContext::
+environmentVariables() const
```
Ktwu: I imagine a map could be pretty hefty to copy/return -- maybe return a shared pointer instead?

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    return environment;
+}
+
+std::string DefaultProcessContext::
+userName() const
+{
+    std::string result;
+
+    if (struct passwd const *pw = ::getpwuid(::getuid())) {
```
Ktwu: What's the difference between getpwuid() and getpwuent()?  A quick lookup for me indicates you should call endpwent() for the latter case.

http://pubs.opengroup.org/onlinepubs/009695399/functions/endpwent.html

Libraries/libutil/Headers/libutil/MemoryProcessContext.h
```diff
@@ -0,0 +1,89 @@
...
+        int32_t groupID,
+        std::string const &userName,
+        std::string const &groupName);
+    explicit MemoryProcessContext(ProcessContext const &processContext);
+    virtual ~MemoryProcessContext();
+
+public:
+    virtual std::string executablePath() const
+    { return _executablePath; }
+    std::string &executablePath()
```
Ktwu: Make this also const?  Same request for all other reference getters.

Libraries/libutil/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_library(util SHARED
             Sources/MemoryFilesystem.cpp
```
Ktwu: Other than const nits, LGTM.

Libraries/libutil/Sources/DefaultProcessContext.cpp
```diff
@@ -112,6 +112,29 @@ executablePath() const
 #endif
 }
 
+static int commandLineArgumentCount = 0;
```
Ktwu: Rename these to make it clear they're static arguments, like sCommandLineArgumentCount?

Libraries/libutil/Sources/MemoryProcessContext.cpp
```diff
@@ -32,15 +34,16 @@ MemoryProcessContext(
 }
 
 MemoryProcessContext::
-MemoryProcessContext(ProcessContext const &processContext) :
+MemoryProcessContext(ProcessContext const *processContext) :
```
Ktwu: Why &->* ?

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Hmm... xcbuild doesn't have prefixes like that anywhere else, but also it's useful to know.

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: 😄

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Just moved this from `SysUtil` here, but could do a refactor pass after.

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Just moved this from `SysUtil` here, but could do a refactor pass after.

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Yeah, I wanted these to return references, but that's hard to do with globals since they need to be initialised. I'll do a follow-up to change that after this lands?

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: This is just moved from `SysUtil` which I didn't write, so not sure what the original thinking was, so nice find! Will remove in a follow-up?

Libraries/libutil/Headers/libutil/MemoryProcessContext.h
grp: The mutable references here are mostly to be setters. The getters above should be `const &`, but that was harder to implement for `DefaultProcessContext` and has to be the same for both.

Libraries/libutil/Sources/MemoryProcessContext.cpp
grp: These are usually passed around by pointer since they have virtual methods, and it's non-obvious that calling via references goes through the vtable. Same reasoning here, just forgot that at first.

Libraries/libutil/Sources/DefaultProcessContext.cpp
Ktwu: Followup is fine.

Libraries/libutil/Sources/DefaultProcessContext.cpp
Ktwu: Yah that's fine.

Libraries/libutil/Sources/MemoryProcessContext.cpp
Ktwu: Mk.

Libraries/libutil/Headers/libutil/DefaultProcessContext.h
```diff
@@ -19,14 +19,15 @@ namespace libutil {
  */
 class DefaultProcessContext : public ProcessContext {
```
Ktwu: LGTM since I only commented with nits.

Libraries/libutil/Headers/libutil/ProcessContext.h
```diff
@@ -85,7 +85,7 @@ class ProcessContext {
     /*
      * Get the system instance.
```
Ktwu: Maybe add a note that the caller should strive to use a context passed from above whenever possible.

Libraries/pbxbuild/Sources/Tool/Invocation.cpp
```diff
@@ -115,7 +115,7 @@ Absolute(std::string const &path)
 Executable Executable::
 Builtin(std::string const &name)
 {
-    std::string executableRoot = FSUtil::GetDirectoryName(ProcessContext::GetDefault()->executablePath());
+    std::string executableRoot = FSUtil::GetDirectoryName(ProcessContext::GetDefaultUNSAFE()->executablePath());
```
Ktwu: Any plans to try and pass in the context from above?

Libraries/pbxproj/Tools/dump_xcodeproj.cpp
```diff
@@ -414,7 +416,7 @@ main(int argc, char **argv)
 
     //fprintf(stderr, "parse ok\n");
```
Ktwu: Get rid of this comment while you're here?

Libraries/acdriver/Headers/acdriver/Driver.h
```diff
@@ -14,6 +14,7 @@
 #include <vector>
```
Ktwu: I really like this change!  LGTM!

Can you explain what's stopping you from adding the process context everywhere (like, is it just too messy to pipe the context through to pbxbuild, or is it unnecessary)?

Libraries/pbxbuild/Headers/pbxbuild/Build/Environment.h
```diff
@@ -17,6 +17,7 @@
 #include <ext/optional>
```
Ktwu: LGTM++

Libraries/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ add_subdirectory(pbxproj)
 add_subdirectory(pbxsetting)
```
Ktwu: Each process-related commit seems better than the next.  Yay code cleanup.  LGTM

Ktwu:
Sorry for the delay; went through a few more commits.

Ktwu:
Ugh, I thought my comments would show up per commit, not per file.  Sorry :(

Overall I really like your process refactoring work.

Libraries/builtin/Sources/copy/Driver.cpp
```diff
@@ -39,33 +39,48 @@ name()
...
-    if (!cp.execute(filesystem, "/bin/cp", { "-R", inputPath, outputPath }, environmentVariables, workingDirectory) || cp.exitcode() != 0) {
+    // TODO: This shouldn't use external tools to copy.
+    process::Context const *context = process::Context::GetDefaultUNSAFE();
+    process::Launcher *launcher = process::Launcher::GetDefaultUNSAFE();
+
+    process::MemoryContext cp = process::MemoryContext(
+        "/bin/cp",
+        context->currentDirectory(),
+        { "-R", inputPath, outputPath },
+        context->environmentVariables(),
```
Ktwu: Would it be worth it create another constructor that accepts an exe path, working directory, args, and a context from which it can derive all of these other properties?

Libraries/libutil/Headers/libutil/ProcessContext.h
grp: Good point, will do.

Libraries/pbxbuild/Sources/Tool/Invocation.cpp
grp: This code is called deep inside pbxbuild, which would be hard to pass down. (Same for the filesystem it uses elsewhere.) Instead, I think this needs to essentially record which constructor was called, then the executor would use the process context and filesystem to actually find the executable.

Libraries/acdriver/Headers/acdriver/Driver.h
grp: I want to avoid it in pbxbuild, so it's more usable outside a single process. For example if you built an IDE that used pbxbuild for building, it shouldn't need to spawn another process to run the build.

Libraries/builtin/Sources/copy/Driver.cpp
grp: I originally had this code doing that with setters:

   process::MemoryContext context = process::MemoryContext(processContext);
   context.executablePath() = "/bin/cp";
   context.commandLineArguments() = { "-R", inputPath, outputPath };

But that seemed too easy to forget some options. This way is explicit about which are copied and which are specified.

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Removing in followup.

Libraries/libutil/Sources/DefaultProcessContext.cpp
grp: Done in followup.

Libraries/pbxproj/Tools/dump_xcodeproj.cpp
grp: Done in followup.

PR #128, opened 2016-09-30
Grab specifications from the correct domain

sas:
This had the nasty side effect of breaking with a stack overflow when a `foo:identifier`
xcspec was `BasedOn` a `bar:identifier` xcspec. We were not enforcing the domain,
which meant we kept inheriting from ourselves over and over.

### DIFF
--- a/Libraries/pbxspec/Sources/Manager.cpp
+++ b/Libraries/pbxspec/Sources/Manager.cpp
@@ -460,10 +460,11 @@ Create(void)
     return std::make_shared <Manager> ();
 }
 
-std::string Manager::
+std::string const &Manager::
 AnyDomain()
 {
-    return "<<domain>>";
+    static std::string any = "<<domain>>";
+    return any;
 }
 
 std::vector<std::pair<std::string, std::string>> Manager::
--- a/Libraries/pbxspec/Sources/Manager.cpp
+++ b/Libraries/pbxspec/Sources/Manager.cpp
@@ -285,7 +285,7 @@ addSpecification(PBX::Specification::shared_ptr const &spec)
 }
 
 bool Manager::
-inheritSpecification(PBX::Specification::shared_ptr const &specification)
+inheritSpecification(PBX::Specification::shared_ptr const &specification, std::vector<PBX::Specification::shared_ptr> visited)
 {
     if (specification->basedOnIdentifier() && specification->basedOnDomain() && specification->base() == nullptr) {
         /* Find the base specification. */
@@ -308,8 +308,22 @@ inheritSpecification(PBX::Specification::shared_ptr const &specification)
              * <any>:identifier and potentially find foo:identifier again.
              */
             auto I = std::find_if(candidates.begin(), candidates.end(),
-                                  [&specification](PBX::Specification::shared_ptr const &spec) -> bool {
-                                      return specification != spec;
+                                  [&specification, &visited](PBX::Specification::shared_ptr const &spec) -> bool {
+                                      for (auto duplicate : visited) {
+                                          // check if specification exists in current inheritance chain.
+                                          if (specification == duplicate) {
+                                              return false;
+                                          }
+                                          // check if base candidate already inherited from specification in question.
+                                          auto base = duplicate->base();
+                                          while (base) {
+                                              if (specification == base) {
+                                                  return false;
+                                              }
+                                              base = base->base();
+                                          }
+                                      }
+                                      return true;
                                   });
             if (I != candidates.end()) {
                 base = *I;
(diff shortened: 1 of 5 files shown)


Libraries/pbxspec/Sources/Manager.cpp
```diff
@@ -86,9 +86,13 @@ findSpecification(std::vector<std::string> const &domains, std::string const &id
 {
     typename T::vector vector = findSpecifications <T> (domains, type);
```
grp: Aren't the domains already checked in this call, so don't need to be below?

sas: Aha, looks like you're right. I think the problem comes from `Manager::inheritSpecification`, in the block that says this:

```
        /*
         * Search the specified domain then in any domain. Some specifications inherit
         * from domain/identifier pairs that don't exist in practice, but by using any
         * domain they can successfully inherit from something potentially relevant.
         */
```

I suppose we should do two things:
* make sure that if a domain is specified, and an identifier in that domain exists, we pick that one first instead of picking the one from `<<any domain>>`;
* and then, if we have to rely on the wildcard domain, enforce a check to make sure that we don't inherit from ourselves (which is where the failure I'm trying to fix comes from.

grp: The first point is done already: see the line just below that comment that uses `{ *specification->basedOnDomain(), AnyDomain() }` as the domains, with the explicitly specified domain coming first.

For your second point, I thought about that, but there's also the possibility of an indirect loop. That could be handled by keeping a set of seen (domain, identifier) pairs while recursively resolving the inheritance. Just not written yet.

grp: Another option would be to pull out `pbxbuild::DirectedGraph<T>` and use the toposort from that rather than the current brute force DFS. That has error handling and unit tests, unlike the current specification inheritance code. 😭

sas: Yeah. That makes sense. In the meantime what do you recommend? Checking this in to at least fix the easy case of the bug? I'm not familiar with the `pbxbuild::DirectedGraph` code yet, but we should probably file an issue for that and take a look at pulling it out at some point.

Libraries/pbxspec/Sources/Manager.cpp
```diff
@@ -460,10 +511,11 @@ Create(void)
     return std::make_shared <Manager> ();
 }
 
-std::string Manager::
+std::string const &Manager::
 AnyDomain()
 {
-    return "<<domain>>";
+    static std::string any = "<<domain>>";
```
sas: Depends on what you mean by "bad". This object is going to be constructed the first time you call the function, which seems fine to me?

Libraries/pbxspec/Sources/PBX/Specification.cpp
```diff
@@ -38,6 +38,18 @@ Specification()
 }
 
 bool Specification::
+operator==(Specification const &rhs) const
+{
+    return _identifier == rhs._identifier && _domain == rhs._domain;
```
sas: My understanding is that if we do this right, we shouldn't have any duplicate identifier/domain couples for specifications, so effectively, comparing the ID compares the spec itself.

That being said, I'm fine with not using `operator==` for this, but I've got really no idea what the method should be named, so I'll take suggestions.

Libraries/pbxspec/Sources/Manager.cpp
```diff
@@ -307,16 +338,35 @@ bool Manager::
...
+            }
+        }
 
-        /* Find the base specification. */
-        auto base = this->specification(specification->type(), *specification->basedOnIdentifier(), domains);
-        if (base == nullptr) {
+        if (!base) {
             fprintf(stderr, "error: cannot find base %s specification '%s:%s'\n", specification->type(), specification->basedOnDomain()->c_str(), specification->basedOnIdentifier()->c_str());
             return false;
         }
```
grp: AFAICT, this commit would be equivalent to these lines right after this one:

    if (base == specification) {
        fprintf(stderr, "error: cannot inherit specification from itself\n");
        return false;
    }

The pointer equality is fine since at this point `specification` and `base` come from `_specifications`, which had duplicates checked from `addSpecification()`.

It doesn't longer cycles (e.g. `com.example.a` inherits from `com.example.b` and `com.example.b` inherits from `com.example.a`) but I don't think your changes above do either?

grp:
This has two major issues:

1. It only handles the direct self-inheritance case. Instead it should handle cycles of any length. The easiest way to do that would be to reuse the topological sort from the build code which detects cycles, rather than the current hardcoded depth-first search.

2. Specifications should not inherit from themselves. If that's happening it means your specifications are wrong: you may be missing one or something. This PR can help with debug output for that case, but it doesn't actually fix your underlying problem.

Finally, there's also the minor issue of the comparison operator being defined strangely as per my above comment.

But per (2) I think it's not useful to merge this in the incomplete (1) state since it's only use would be to help with debugging specifications.

modocache:
Oh, I see. Yes, you're right, it was a problem in my specifications; fixing the underlying problem allowed `xcbuild` to build my local project just fine. Thanks!

sas:
Sweet.

PR #129, opened 2016-09-30
Installation fixes

Ktwu:
I can't install xcbuild because of these specification errors.  I've no idea what they should actually be named, but this works for me.


Ktwu:
Welp.

PR #130, opened 2016-09-30
Fix linux builds with linenoise

sas:
A couple of things were broken with linux builds and linenoise. This PR fixes them.


grp:
LGTM but conflicts.

PR #131, opened 2016-09-30
Make the unit tests ASAN-clean and run on Travis

sas:


CMakeLists.txt
```diff
@@ -28,8 +28,24 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 # Unused language features.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fno-exceptions")
 
-#set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # Global options (these apply for other subprojects like JSObjects) and must
```
grp: I think this is from ds2, but doesn't apply here.

CMakeLists.txt
```diff
@@ -28,8 +28,24 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
...
+  # be set before we include subdirectories.
+  if (SANITIZER)
+    set(COMMON_FLAGS "${COMMON_FLAGS} -g -fno-omit-frame-pointer")
+  endif ()
+
+  if (SANITIZER STREQUAL "asan")
```
grp: Why use "asan" here? With `-DSANITIZE=address` it could just be `-fsanitize=${SANITIZE}`, much less code. Or even just pass it via `CMAKE_CXX_FLAGS` directly?

CMakeLists.txt
```diff
@@ -28,8 +28,24 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 # Unused language features.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -fno-exceptions")
 
-#set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # Global options (these apply for other subprojects like JSObjects) and must
+  # be set before we include subdirectories.
+  if (SANITIZER)
+    set(COMMON_FLAGS "${COMMON_FLAGS} -g -fno-omit-frame-pointer")
```
grp: Could this be replaced by `-DCMAKE_BUILD_TYPE=Debug` in the Makefile?

CMakeLists.txt
sas: Correct. Let me fix this.

CMakeLists.txt
sas: We do it this way because "asan" is the name people know about, and because for other sanitizers (e.g.: ubsan), we can't just take the cmake argument and pass it down.

CMakeLists.txt
sas: I want the cmake to be usable standalone. Also, ASAN != debug; you could build release (with optimizations) and have asan enabled and generate debug info to print backtraces.

grp:
Weird error:

    /usr/bin/ld: ThirdParty/googletest/googletest/libgtest.a(gtest-all.cc.o): undefined reference to symbol 'pthread_key_delete@@GLIBC_2.2.5'
    //lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line

Also, is there a way that Travis could check all of the sanitizers at once?

grp:
(Google suggests the error means a missing `-lpthread` on the command line?)

PR #133, opened 2016-10-03
Regenerate when the xcbuild binary changes

sas:
If the generator itself (xcbuild) changes, we should regenerate the
ninja files too.


Ktwu:
Makes sense.  LGTM.

PR #135, opened 2016-10-04
Use the dependency info tool only when copying directories

sas:
If we're just copying a standard file, there's no point in using this,
and if we do, the tool complains that the path passed to the
`directory:` argument is not a directory. We have to specify these paths
as simple input files instead.

### DIFF
--- a/Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
+++ b/Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
@@ -13,6 +13,7 @@
 #include <pbxbuild/Tool/Tokens.h>
 #include <pbxbuild/Tool/Context.h>
 #include <libutil/Filesystem.h>
+#include <libutil/DefaultFilesystem.h>
 #include <libutil/FSUtil.h>
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


Libraries/pbxbuild/Sources/Tool/CopyResolver.cpp
```diff
@@ -77,13 +78,20 @@ ResolveInternal(
...
+    std::vector<std::string> toolInputPaths = toolEnvironment.inputs(toolContext->workingDirectory());
+
     // TODO(grp): This should be generic for all tools.
     std::vector<Tool::Invocation::DependencyInfo> dependencyInfo;
     if (tool->deeplyStatInputDirectories()) {
         for (std::string const &inputPath : inputPaths) {
             /* Create a dependency info file to track the input directory contents. */
-            auto info = Tool::Invocation::DependencyInfo(dependency::DependencyInfoFormat::Directory, inputPath);
-            dependencyInfo.push_back(info);
+            libutil::DefaultFilesystem filesystem = libutil::DefaultFilesystem();
```
grp: This is bad, it makes this tool not self-contained. If this is needed here, use `Filesystem::GetDefaultUNSAFE()` to indicate it needs to be replaced by passing down a filesystem.

On the other hand, I think it'd be better to change the dependency info tool to ignore files rather than error. Then this code doesn't have to read the filesystem at all. (I'd prefer if filesystem reads are only done during execution, otherwise Ninja generation could be wrong if this isn't added to the regeneration rule.)

sas: I think that makes sense, but if the dependency tool ignores files, then the dependency is going to be broken, no? Maybe instead of ignoring files, the tool should actually handle them properly and output info for a single file?

grp: It should be fine since the inputs are listed explicitly for Ninja. The directory dependency info tool is to record the contents of the directory, so Ninja notices if directory contents change, rather than just if the directory itself does. For files that's not important.

grp:
In favor of #215.

PR #136, opened 2016-10-04
[Rendition] Make sure 'algorithm' is initialized for release builds

Ktwu:
When build xcbuild with -NDEBUG, assert() is compiled out and, to the compiler,
the 'algorithm' variable has the potential to be uninitialized. Refactor the code
so it compiles cleanly.

### DIFF
--- a/Libraries/libcar/Sources/Rendition.cpp
+++ b/Libraries/libcar/Sources/Rendition.cpp
@@ -377,14 +377,9 @@ Decode(struct car_rendition_value *value)
             return ext::nullopt;
         } else if (header1->compression == car_rendition_data_compression_magic_lzvn || header1->compression == car_rendition_data_compression_magic_jpeg_lzfse) {
 #if HAVE_LIBCOMPRESSION
-            compression_algorithm algorithm;
-            if (header1->compression == car_rendition_data_compression_magic_lzvn) {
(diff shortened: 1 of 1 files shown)


Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -377,14 +377,9 @@ Decode(struct car_rendition_value *value)
...
-            if (header1->compression == car_rendition_data_compression_magic_lzvn) {
-                algorithm = (compression_algorithm)_COMPRESSION_LZVN;
-            } else if (header1->compression == car_rendition_data_compression_magic_jpeg_lzfse) {
-                algorithm = COMPRESSION_LZFSE;
-            } else {
-                assert(false);
-            }
+            compression_algorithm algorithm = header1->compression == car_rendition_data_compression_magic_lzvn ?
+                (compression_algorithm)_COMPRESSION_LZVN :
+                COMPRESSION_LZFSE;
```
sas: Instead of defaulting to `COMPRESSION_LZFSE`, shouldn't we return an error or something? It seems like with your code we're gonna use LZFSE for everything other than LZVM and LZFSE, which might be a bug.

Ktwu: This chunk of code should already be guarded by a check for these two types on line 378.  There should be no reasonable reason for the compression type to be anything else.

PR #137, opened 2016-10-05
New asset types and properties.

grp:
There's still a few more to add, but these ones should be ready as-is.

### DIFF
--- a/Libraries/acdriver/Sources/Compile/StickerSequence.cpp
+++ b/Libraries/acdriver/Sources/Compile/StickerSequence.cpp
@@ -1,0 +1,35 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <acdriver/Compile/StickerSequence.h>
+#include <acdriver/Compile/Output.h>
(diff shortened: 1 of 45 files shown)


Libraries/xcassets/Sources/Slot/ColorSpace.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <cstdlib>
+
+using xcassets::Slot::ColorSpace;
+using xcassets::Slot::ColorSpaces;
+
+ext::optional<ColorSpace> ColorSpaces::
+Parse(std::string const &value)
+{
+    if (value == "sRGB") {
```
sas: Having long enums like this is really not the nicest way of handling these mappings :/

Libraries/acdriver/Sources/Compile/StickerSequence.cpp
```diff
@@ -0,0 +1,35 @@
...
+    Filesystem *filesystem,
+    Output *compileOutput,
+    Result *result)
+{
+    result->document(
+        Result::Severity::Warning,
+        stickerSequence->path(),
+        { Output::AssetReference(stickerSequence) },
+        "Not Implemented",
+        "sticker sequence yet supported");
```
sas: I think you meant `**not** yet supported` on these.

Libraries/xcassets/Sources/Slot/ColorSpace.cpp
grp: Long enums?

PR #142, opened 2016-10-06
[actool] Add --compile-output-filename option for specifying car filename

kolinkrewinkel:
Having the option to customize the output path fully (not just the directory) is helpful in cases where the archives are being pre-generated, or there are multiple going into the same directory.

### DIFF
--- a/Libraries/acdriver/Sources/CompileAction.cpp
+++ b/Libraries/acdriver/Sources/CompileAction.cpp
@@ -262,7 +262,8 @@ run(Filesystem *filesystem, Options const &options, Output *output, Result *resu
      * If necessary, create output archive to write into.
      */
     if (compileOutput.format() == Compile::Output::Format::Compiled) {
-        std::string path = compileOutput.root() + "/" + "Assets.car";
+        std::string outputFilename = options.compileOutputFilename().value_or("Assets.car");
+        std::string path = compileOutput.root() + "/" + outputFilename;
 
         ext::optional<car::Writer> writer = CreateWriter(path);
         if (!writer) {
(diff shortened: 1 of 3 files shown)


Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -262,7 +262,8 @@ run(Filesystem *filesystem, Options const &options, Output *output, Result *resu
      * If necessary, create output archive to write into.
      */
     if (compileOutput.format() == Compile::Output::Format::Compiled) {
-        std::string path = compileOutput.root() + "/" + "Assets.car";
+        const char *outputFilename = options.compileOutputFilename().value_or("Assets.car").c_str();
```
grp: Why `.c_str()` here? It's going back to a `std::string` on the next line?

Libraries/acdriver/Sources/Options.cpp
```diff
@@ -32,6 +32,8 @@ parseArgument(std::vector<std::string> const &args, std::vector<std::string>::co
         return libutil::Options::Current<bool>(&_printContents, arg);
     } else if (arg == "--compile") {
         return libutil::Options::Next<std::string>(&_compile, args, it);
+    } else if (arg == "--compile-output-filename") {
```
grp: Is this an extension or a new option?

sas: Can we call this just `--output` or is there a reason why the option has to be that long? Does this tool generate multiple output files?

Libraries/acdriver/Sources/CompileAction.cpp
kolinkrewinkel: Whoops, fixed in the updated commit

Libraries/acdriver/Sources/Options.cpp
kolinkrewinkel: Extension. I made it long (including compile output) because `--compile` includes both the source and the destination directory, and I wanted to be clear that this will affect the results of things (the `output`/destination) specifically when you use `compile`

grp: Could you add a comment somewhere (maybe in the options struct, like `Libraries/xcdriver/Headers/xcdriver/Options.h`) that it's an extension? Makes it easier to keep track of.

kolinkrewinkel: Done in 05ac2e5!

Issue #138, opened 2016-10-06
__attribute__((constructor)) is unreliable for grabbing argv/argc


Ktwu:
My Linux build of plutil (and I assume other tools) fails to properly grab argv/argc in DefaultContext, which results in a crash later on in the code -- the arguments passed to these constructors are nil.

PR #146, opened 2016-10-07
[process] Don't use the constructor attribute on Linux for grabbing command line arguments

Ktwu:
h/t to @sas for this fix. This is a minimal fix for my problems when
using xcbuild tools on my Linux machine.

See https://github.com/facebook/xcbuild/issues/138

### DIFF
--- a/Libraries/process/Sources/DefaultContext.cpp
+++ b/Libraries/process/Sources/DefaultContext.cpp
@@ -128,12 +128,18 @@ executablePath() const
 static int commandLineArgumentCount = 0;
 static char **commandLineArgumentValues = NULL;
 
+#if !defined(__linux__)
 __attribute__((constructor))
+#endif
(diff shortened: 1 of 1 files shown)


Libraries/process/Sources/DefaultContext.cpp
```diff
@@ -128,13 +128,19 @@ executablePath() const
...
 __attribute__((constructor))
+#endif
 static void CommandLineArgumentsInitialize(int argc, char **argv)
 {
     commandLineArgumentCount = argc;
     commandLineArgumentValues = argv;
 }
 
+#if defined(__linux__)
+__attribute__((section(".init_array"))) void *construct_it = (void*)&CommandLineArgumentsInitialize;
```
Ktwu: Done.

grp:
Curious about the history that led to two sections for constructor functions with different behavior.

PR #147, opened 2016-10-08
Lots more changes to asset catalog loading.

grp:
- Parse and load remaining types and almost all remaining properties.
 - Fix bugs, including completely ignoring the previously added sticker types.
 - Change child asset loading behavior to simplify & improve accessing child assets.
 - Other cleanup to asset loading code.

Next up for this code is probably handling errors correctly, rather than logging to `stderr` all the time.


grp:
Ah, good to know. Thanks.

jackwu95:
Yep, I'll try my best to help out while he's gone :)

grp:
@jackwu95 you might need to add yourself to the Facebook GitHub organization for your accept to work?

PR #148, opened 2016-10-08
Allocate enough space for Unicode strings in binary property lists.

### LINKED ISSUE #145: xcbuild Intermittent Fatal malloc() Error
On my system (macOS 10.12), `xcbuild` (or the `xcodebuild` tool installed by it, which I assume is synonymous) crashes with a fatal error deep within the system `malloc` library. This crash is intermittent (sometimes xcbuild launches just fine), but when it does happen, the crash is immediate. [Here is a sample crash log.](https://gist.github.com/wjk/603be9f548a308207e95ac4451c43402) Is there anything that can be done about this? I need to use `xcbuild` for my project because of its support for custom specification files (Apple `xcodebuild` does not support this as far as I know), and this crash is stopping me from being able to continue. Thanks!

grp:
Before, only half as much space as needed was allocated. Fixes #145.


grp:
This seems right and it makes parsing get (much) further, but I don't know this code that well.

Libraries/plist/Sources/Format/Binary.cpp
```diff
@@ -176,7 +176,7 @@ Create(void *opaque, ABPRecordType type, void *arg1, void *arg2, void *arg3)
                 if (ReadSeek(opaque, offset, SEEK_SET) < 0)
                     return nullptr;
 
-                string.resize(nchars);
+                string.resize(sizeof(char) * nchars);
```
Ktwu: Nit: both here and for UTF16, use something like 'size_t stringsize = sizeof(foo) * nchar'

grp:
I'll add a test, although I'm not sure how to check anything except that it doesn't violate ASAN. With #131 that should be enough though.

grp:
Test added. Turns out it didn't decode the right string before either so the test did fail before the fix.

Ktwu:
Lgtm

Ktwu:
Sorry, finger slipped in mobile ui.  Merge at your leisure.

PR #149, opened 2016-10-08
Add definitions for some newer asset archive format fields.

grp:
The extra metadata section stores a string with contextual info
on how the asset archive was compiled, especially when the archive
is stripped to target certain devices or platforms.

Other new fields correspond to new asset document format fields.

### DIFF
--- a/Libraries/libcar/Headers/car/car_format.h
+++ b/Libraries/libcar/Headers/car/car_format.h
@@ -108,9 +108,11 @@ enum car_attribute_identifier {
     car_attribute_identifier_size_class_vertical = 21,
     car_attribute_identifier_memory_class = 22,
     car_attribute_identifier_graphics_class = 23,
+    car_attribute_identifier_display_gamut = 24,
+    car_attribute_identifier_deployment_target = 25,
 
     /* Not a real value; used as a marker for the maximum identifier. */
-    _car_attribute_identifier_count = 24,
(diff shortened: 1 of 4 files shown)


Libraries/libcar/Headers/car/car_format.h
```diff
@@ -196,6 +198,11 @@ struct car_header {
     uint32_t key_semantics; // 01 00 00 00
 } __attribute__((packed));
 
+struct car_extended_metadata {
+    char magic[4]; // 'META'
+    char contents[1024]; // string
```
Ktwu: I'm guessing 1024 characters is a known max for extended metadata?

grp: Hmm, I thought it was a fixed-length format since there was no size field, but I realize now that that could be computed from the BOM index size containing the extended metadata. I think fixed-size makes more sense, though...

Libraries/libcar/Headers/car/car_format.h
```diff
@@ -196,6 +198,11 @@ struct car_header {
     uint32_t key_semantics; // 01 00 00 00
 } __attribute__((packed));
 
+struct car_extended_metadata {
+    char magic[4]; // 'META'
```
jackwu95: The endianness might be an issue here again, add a todo to keep that in mind?

PR #151, opened 2016-10-11
[actool] Fix car header key format in car writer

jackwu95:
Looks like the key format is also little-endian.

Fixes the issue:
```
CoreUI:  Car file has erroneous key format information. Using CUISystemThemeRenditionKeyFormat
```
as well as the subsequent buffer overflow when reading the compiled car file.


grp:
On a related note, `libcar` needs an endianness audit. Most of it relies on the host system being little-endian.

PR #155, opened 2016-10-12
Modernize and complete SDK loading.

grp:
- Warn about missing keys when parsing; add currently missing keys.
 - Use optionals to avoid unsafely assuming parsed values always exist.
 - Update style to match other libraries.

### DIFF
--- a/Libraries/builtin/Sources/infoPlistUtility/Driver.cpp
+++ b/Libraries/builtin/Sources/infoPlistUtility/Driver.cpp
@@ -97,6 +97,8 @@ ExpandBuildSettings(plist::Object *value, pbxsetting::Environment const &environ
 static void
 AddBuildEnvironment(plist::Dictionary *root, pbxsetting::Environment const &environment)
 {
+    // TODO: This should use `xcsdk::SDK::Platform::additionalInfo()`.
+
     root->set("DTCompiler", plist::String::New(environment.resolve("DEFAULT_COMPILER")));
     root->set("DTXcode", plist::String::New(environment.resolve("XCODE_VERSION_ACTUAL")));
     root->set("DTXcodeBuild", plist::String::New(environment.resolve("XCODE_PRODUCT_BUILD_VERSION")));
--- a/Libraries/pbxbuild/Sources/Target/Environment.cpp
+++ b/Libraries/pbxbuild/Sources/Target/Environment.cpp
@@ -384,13 +384,21 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
         pbxsetting::Setting::Parse("GCC_VERSION", "$(DEFAULT_COMPILER)"),
     }), false);
 
-    environment.insertFront(sdk->platform()->defaultProperties(), false);
+    if (sdk->platform()->defaultProperties()) {
+        environment.insertFront(*sdk->platform()->defaultProperties(), false);
+    }
     environment.insertFront(PlatformArchitecturesLevel(buildEnvironment.specManager(), specDomains), false);
-    environment.insertFront(sdk->defaultProperties(), false);
+    if (sdk->defaultProperties()) {
+        environment.insertFront(*sdk->defaultProperties(), false);
+    }
     environment.insertFront(sdk->platform()->settings(), false);
     environment.insertFront(sdk->settings(), false);
-    environment.insertFront(sdk->customProperties(), false);
-    environment.insertFront(sdk->platform()->overrideProperties(), false);
+    if (sdk->customProperties()) {
+        environment.insertFront(*sdk->customProperties(), false);
+    }
+    if (sdk->platform()->overrideProperties()) {
+        environment.insertFront(*sdk->platform()->overrideProperties(), false);
+    }
 
     if (packageType != nullptr) {
         environment.insertFront(PackageTypeLevel(packageType), false);
@@ -435,6 +443,7 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     std::vector<xcsdk::SDK::Toolchain::shared_ptr> toolchains;
     for (std::string const &toolchainName : pbxsetting::Type::ParseList(environment.resolve("TOOLCHAINS"))) {
         if (xcsdk::SDK::Toolchain::shared_ptr toolchain = buildEnvironment.sdkManager()->findToolchain(toolchainName)) {
+            // TODO: Apply toolchain override build settings.
             toolchains.push_back(toolchain);
         }
     }
--- a/Libraries/xcsdk/Sources/SDK/Platform.cpp
+++ b/Libraries/xcsdk/Sources/SDK/Platform.cpp
@@ -12,9 +12,12 @@
 #include <pbxsetting/Type.h>
 #include <libutil/Filesystem.h>
 #include <libutil/FSUtil.h>
+#include <plist/Array.h>
+#include <plist/Boolean.h>
 #include <plist/Dictionary.h>
 #include <plist/String.h>
 #include <plist/Format/Any.h>
+#include <plist/Keys/Unpack.h>
 
 #include <algorithm>
 
@@ -24,9 +27,7 @@ using libutil::FSUtil;
 
 Platform::
 Platform() :
-    _defaultDebuggerSettings(nullptr),
-    _defaultProperties      (pbxsetting::Level({ })),
-    _overrideProperties     (pbxsetting::Level({ }))
+    _defaultDebuggerSettings(nullptr)
 {
 }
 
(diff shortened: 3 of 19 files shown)


Libraries/xcsdk/Sources/SDK/Platform.cpp
```diff
@@ -63,7 +64,7 @@ settings() const
 {
     std::vector<pbxsetting::Setting> settings = {
         pbxsetting::Setting::Create("PLATFORM_NAME", _name),
-        pbxsetting::Setting::Create("PLATFORM_DISPLAY_NAME", _description),
+        pbxsetting::Setting::Create("PLATFORM_DISPLAY_NAME", _description.value_or("")),
```
Ktwu: What would happen if you just didn't include this particular setting if _description didn't exist?

Libraries/xcsdk/Sources/SDK/Platform.cpp
```diff
@@ -155,17 +156,46 @@ executablePaths() const
...
+    (void)unpack.cast <plist::String> ("DTSDKName");
+    (void)unpack.cast <plist::String> ("DTPlatformBuild");
+    (void)unpack.cast <plist::String> ("DTPlatformVersion");
+    (void)unpack.cast <plist::String> ("DTCompiler");
+    (void)unpack.cast <plist::String> ("DTXcode");
+    (void)unpack.cast <plist::String> ("DTXcodeBuild");
+    (void)unpack.cast <plist::String> ("BuildMachineOSBuild");
+
+    if (!unpack.complete(true)) {
+        fprintf(stderr, "%s", unpack.errorText().c_str());
```
Ktwu: return false?

Libraries/xcsdk/Sources/SDK/Product.cpp
```diff
@@ -12,22 +12,31 @@
...
+    auto unpack = plist::Keys::Unpack("Product", dict, &seen);
+
+    auto PN   = unpack.cast <plist::String> ("ProductName");
+    auto PV   = unpack.cast <plist::String> ("ProductVersion");
+    auto PUVV = unpack.cast <plist::String> ("ProductUserVisibleVersion");
+    auto PBV  = unpack.cast <plist::String> ("ProductBuildVersion");
+    auto PC   = unpack.cast <plist::String> ("ProductCopyright");
+
+    if (!unpack.complete(true)) {
+        fprintf(stderr, "%s", unpack.errorText().c_str());
```
Ktwu: Once again:  return false?

Libraries/xcsdk/Sources/SDK/Target.cpp
```diff
@@ -61,19 +63,50 @@ executablePaths() const
...
+    auto DP    = unpack.cast <plist::Dictionary> ("DefaultProperties");
+    auto CP    = unpack.cast <plist::Dictionary> ("CustomProperties");
+    auto PCFNs = unpack.cast <plist::Array> ("PropertyConditionFallbackNames");
+    auto DSFN  = unpack.cast <plist::String> ("DocSetFeedName");
+    auto DSFU  = unpack.cast <plist::String> ("DocSetFeedURL");
+    auto MSTV  = unpack.cast <plist::String> ("MinimumSupportedToolsVersion");
+    auto SBTCs = unpack.cast <plist::Array> ("SupportedBuildToolComponents");
+
+    /* Ignored: seems to be a typo. */
+    (void)unpack.cast <plist::String> ("isBaseSDK");
```
Ktwu: Why all of these ignored unpackings?  I assume unpack.complete() won't return true otherwise?

Libraries/xcsdk/Sources/SDK/Target.cpp
```diff
@@ -61,19 +63,50 @@ executablePaths() const
...
-        _version = V->value();
+    if (Ts  != nullptr) {
+        if (std::shared_ptr<Manager> manager = _manager.lock()) {
+            for (size_t n = 0; n < Ts ->count(); n++) {
+                if (auto T = Ts ->value <plist::String> (n)) {
+                    if (auto toolchain = manager->findToolchain(T->value())) {
+                        _toolchains.push_back(toolchain);
+                    } else if (auto defaultToolchain = manager->findToolchain(Toolchain::DefaultIdentifier())) {
+                        _toolchains.push_back(defaultToolchain);
+                    }
```
Ktwu: Error if no toolchain found?

Libraries/xcsdk/Sources/SDK/Target.cpp
```diff
@@ -61,19 +63,50 @@ executablePaths() const
...
+                }
+            }
+        }
+    } else {
+        if (std::shared_ptr<Manager> manager = _manager.lock()) {
+            if (auto defaultToolchain = manager->findToolchain(Toolchain::DefaultIdentifier())) {
+                _toolchains.push_back(defaultToolchain);
+            }
```
Ktwu: Error if no default toolchain?

Libraries/xcsdk/Tests/test_PlatformVersion.cpp
```diff
@@ -37,8 +37,8 @@ TEST(PlatformVersion, Parse)
     auto platformVersion = PlatformVersion::Open(&filesystem, "/" + platform);
```
Ktwu: For all the new properties you've added parsing support for, can you write a quest test to make sure they're represented?

Libraries/builtin/Sources/infoPlistUtility/Driver.cpp
```diff
@@ -97,6 +97,8 @@ ExpandBuildSettings(plist::Object *value, pbxsetting::Environment const &environ
 static void
 AddBuildEnvironment(plist::Dictionary *root, pbxsetting::Environment const &environment)
 {
+    // TODO: This should use `xcsdk::SDK::Platform::additionalInfo()`.
```
grp: Not important enough for an issue IMO, doesn't affect behavior at all.

Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -384,13 +384,21 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
...
-    environment.insertFront(sdk->platform()->defaultProperties(), false);
+    if (sdk->platform()->defaultProperties()) {
+        environment.insertFront(*sdk->platform()->defaultProperties(), false);
+    }
     environment.insertFront(PlatformArchitecturesLevel(buildEnvironment.specManager(), specDomains), false);
-    environment.insertFront(sdk->defaultProperties(), false);
+    if (sdk->defaultProperties()) {
+        environment.insertFront(*sdk->defaultProperties(), false);
+    }
     environment.insertFront(sdk->platform()->settings(), false);
```
grp: With this change, `defaultProperties()` becomes an `ext::optional<pbxsetting::Level>`, which means it may not exist. (Depending on whether it existed in the original file.) `settings()` is not optional, since it's synthesized at runtime, so nothing to check for.

The advantage of storing optionals is that it maintains the information of whether the source file had an empty `defaultProperties()` (optional has value, value is empty), or didn't have one at all (optional has no value).

Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -435,6 +443,7 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     std::vector<xcsdk::SDK::Toolchain::shared_ptr> toolchains;
     for (std::string const &toolchainName : pbxsetting::Type::ParseList(environment.resolve("TOOLCHAINS"))) {
         if (xcsdk::SDK::Toolchain::shared_ptr toolchain = buildEnvironment.sdkManager()->findToolchain(toolchainName)) {
+            // TODO: Apply toolchain override build settings.
```
grp: I don't think this does anything in practice, so not worth an issue yet IMO.

Libraries/xcsdk/Sources/SDK/Platform.cpp
grp: Same, this is just warnings from unknown keys.

Libraries/xcsdk/Sources/SDK/Product.cpp
grp: Same. These are warnings for unknown keys, which isn't an error.

Libraries/xcsdk/Sources/SDK/Target.cpp
grp: That's fine, if during a build one is needed and not found it'll error out. But even if none is found here, someone could specify `TOOLCHAINS` on the command line and still build.

Libraries/xcsdk/Tests/test_PlatformVersion.cpp
grp: In general, I think it's too much work and not too valuable to test parsing each key in each type of thing that's added. There's thousands of keys and it's just checking that they're written right in the original file. My current plan is to add tests for any keys that need *changes* to their parsing, usually from a bug.

Ktwu:
LGTM since you responded to my nittiness.   What do you think about xcbuild reporting if it parses a key that it currently doesn't do anything with?

Libraries/xcsdk/Sources/SDK/Platform.cpp
grp: No change, but it wouldn't show in `-showBuildSettings` like it should.

grp:
> What do you think about xcbuild reporting if it parses a key that it currently doesn't do anything with?

Generally I think the code that uses the keys would do that if they're important. Many of these keys are UI-related, for example, which xcbuild would never use.

PR #162, opened 2016-10-18
Minor cleanups to previous commit.

grp:
- Use (const) pointers instead of references to match existing style.
 - Replace switches on types with casting, reduce switch nesting.

### DIFF
--- a/Libraries/plist/Tools/PlistBuddy.cpp
+++ b/Libraries/plist/Tools/PlistBuddy.cpp
@@ -88,40 +88,6 @@ Options::
 {
 }
 
-static std::unique_ptr<plist::Object>
-ReadPlistObjectFromFile(Filesystem &filesystem, std::string filepath)
-{
-    if (filepath.empty()) {
-        fprintf(stderr, "no input specified\n");
-        return NULL;
-    }
-
-    if (!filesystem.exists(filepath)) {
-        fprintf(stderr, "File does not exist: %s\n", filepath.c_str());
-        return NULL;
-    }
-
-    std::vector<uint8_t> contents;
-    if (!filesystem.read(&contents, filepath)) {
-        fprintf(stderr, "error: unable to read %s\n", filepath.c_str());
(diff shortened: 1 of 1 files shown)


Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -206,6 +172,34 @@ CommandHelp()
...
+        return NULL;
+    }
+
+    std::vector<uint8_t> contents;
+    if (!filesystem->read(&contents, path)) {
+        fprintf(stderr, "error: unable to read %s\n", path.c_str());
+        return NULL;
+    }
+
+    auto deserialize = plist::Format::Any::Deserialize(contents);
```
grp: The identified format wasn't being used later, so no need to compute it here: the second argument to `plist::Format::Any::Deserialize` is optional and detects the format automatically by default.

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -467,49 +461,48 @@ Delete(plist::Object *object, std::queue<std::string> *keyPath) {
 }
 
 static bool
-Merge(plist::Object *object, std::string &merge_file, std::queue<std::string> &keyPath, Filesystem *filesystem)
+Merge(plist::Object *object, std::string &mergeSource, std::queue<std::string> *keyPath, Filesystem const *filesystem)
```
Ktwu: Why make keyPath a pointer?  It's not intended to be mutated.

Libraries/plist/Tools/PlistBuddy.cpp
Ktwu: Nice, thanks.

Libraries/plist/Tools/PlistBuddy.cpp
grp: `GetObjectAtKeyPath()` mutates it below, IIRC.

PR #163, opened 2016-10-19
Delay resolving executable paths until execution time.

grp:
Rather than looking at the filesystem to find an executable path at
configuration time, find the executable at execution time. This allows
different executors to handle searching for paths differently (for
example, using internal or external builtin tools), and avoids using
the filesystem during configuration.

### DIFF
--- a/Libraries/xcexecution/Sources/NinjaExecutor.cpp
+++ b/Libraries/xcexecution/Sources/NinjaExecutor.cpp
@@ -114,6 +114,31 @@ NinjaHash(std::string const &input)
     return ss.str();
 }
 
+static ext::optional<std::string>
+NinjaExecutablePath(
+    process::Context const *processContext,
+    Filesystem const *filesystem,
+    std::vector<std::string> const &executablePaths,
+    pbxbuild::Tool::Invocation::Executable const &executable)
+{
+    if (ext::optional<std::string> const &builtin = executable.builtin()) {
+        std::string path = FSUtil::GetDirectoryName(processContext->executablePath()) + "/" + *builtin;
+        if (filesystem->isExecutable(path)) {
+            return path;
+        } else {
+            return ext::nullopt;
+        }
+    } else if (ext::optional<std::string> const &external = executable.external()) {
+        if (FSUtil::IsAbsolutePath(*external)) {
+            return *external;
+        } else {
+            return filesystem->findExecutable(*external, executablePaths);
+        }
+    } else {
+        abort();
+    }
+}
+
 static std::string
 NinjaInvocationPhonyOutput(pbxbuild::Tool::Invocation const &invocation)
 {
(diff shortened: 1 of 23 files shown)


Libraries/xcexecution/Sources/SimpleExecutor.cpp
```diff
@@ -245,39 +244,66 @@ performInvocations(
...
+
+                    xcformatter::Formatter::Print(_formatter->finishInvocation(invocation, *builtin, createProductStructure));
+                } else {
+                    /* Failed to find builtin tool. */
                     return std::make_pair(false, std::vector<pbxbuild::Tool::Invocation>({ invocation }));
                 }
+            } else if (ext::optional<std::string> const &external = executable.external()) {
+                /* External tool, find on the filesystem. */
+                ext::optional<std::string> path;
+                if (FSUtil::IsAbsolutePath(*external)) {
```
grp: Yeah, this is the old `Determine()` logic duplicated. :(

I thought about putting it on `Invocation::Executable` itself, but it would have to take a filesystem parameter and it'd be a bit strange to take that on a pure-data class... maybe could do a shared function in this library?

And there's also the issue that the `builtin` branch is different between the two: in here, it looks it up in the builtin registry, but the Ninja executor uses the filesystem. That makes it a little hard to abstract.

Any ideas?

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -660,10 +697,18 @@ buildTargetInvocations(
...
 
-        /* Write invocations to run after auxiliary files. */
         // TODO(grp): This should perhaps be a separate flag for a 'phony' invocation.
-        if (!invocation.executable().path().empty()) {
-            if (!buildInvocation(&writer, invocation, dependencyInfoToolPath, temporaryDirectory, targetWriteAuxiliaryFiles)) {
+        if (invocation.executable()) {
+            /* Find invocation executable. */
+            ext::optional<std::string> executablePath = NinjaExecutablePath(processContext, filesystem, targetEnvironment.executablePaths(), *invocation.executable());
+            if (!executablePath) {
+                fprintf(stderr, "unable to find executable: %s\n", invocation.executable()->builtin().value_or(invocation.executable()->external().value_or("<NONE>")).c_str());
```
grp: That value is only used in errors like this one — otherwise, it uses the resolved `executablePath` in the log. I avoided adding a method to avoid making the warning message value more obvious than the correct output value.

grp:
Ugh, the "update branch" button makes merge commits. Will have to undo.

Ktwu:
Sorry (and TIL) about 'update branch' :(

grp:
I keep doing it too — it won't let you rebase without it. Which makes no sense the button is to *rebase*, not fast-forward. @sas maybe you can turn it off?

PR #165, opened 2016-10-20
Add json support to plutil.

matthewbauer:
This should add plutil detection.

### DIFF
--- a/Libraries/plist/Headers/plist/Format/Any.h
+++ b/Libraries/plist/Headers/plist/Format/Any.h
@@ -14,6 +14,7 @@
 #include <plist/Format/Binary.h>
 #include <plist/Format/XML.h>
 #include <plist/Format/ASCII.h>
+#include <plist/Format/JSON.h>
 
 namespace plist {
 namespace Format {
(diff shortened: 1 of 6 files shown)


matthewbauer:
I guess I'm not familiar with the Xcode stack enough- but can't we just treat JSON as a *real* plist format? I suspect there's good reason not to but I'm not seeing it. Anyway, I also don't really understand the difference between .xcspec and .plist files.

grp:
It's definitely possible to do treat JSON as a plist format — this PR does it! I can't think of anywhere it'd cause incorrect behavior: the new parsed formats is a superset of the old ones, and invalid data is unlikely to parse as JSON anyway.

My concern is that it makes the code less *precise*. The code says to load a property list, but the implementation might load JSON instead. For me, that makes the code a little more complex and prone to error later on.

About `.xcspec` files: they are property lists, just with a specific structure. It's similar to an `Info.plist`, also a property lists with defined special keys. The `.xcspec` files are usually in the [ASCII (NeXTSTEP)](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/PropertyLists/OldStylePlists/OldStylePLists.html) format, which looks similar to JSON but is unrelated, but I've occasionally them as binary or XML.

copumpkin:
Awesome, thanks! I have no opinion as to the code structure, but I want the feature 😸

PR #164, opened 2016-10-20
Add Specifications

onhachoe:
add basic specifications for
clang analyzer/astbuilder/migrator
coredata compilers (momc, mapc)
strip tool
builtin validation tool

PR #166, opened 2016-10-20
Add new files to CMakeLists.txt

matthewbauer:
Three files were missing: abstreactassetcatalog.xcspec, file.swiftpm-manifest.xcspec, and folder.stickers.xcspec


grp:
Looks like one of these is in #164 too, but this is more complete. Thanks!

PR #167, opened 2016-10-20
Put linker option "-dependency_info $(dependency_info_file)" only on darwin

onhachoe:
linkers in other platforms won't have this option


onhachoe:
Yeah, you have a point, but it is most likely that the target platform linker specification will inherit default domain linker which already have DependencyInfoFile set.
And sadly, AFAIK, there is no way to unset inherited property.
Suggestion on how to solve this?

onhachoe:
Yeah, that sounds like a easy workaround for now.

onhachoe:
Changed to check empty() for dependencyInfoFile

grp:
Actually, I wonder if a linker that doesn't support this flag should be using the `com.apple.pbx.linkers.ld` specification at all, since that one is setup for the Darwin linker with many different flags, not just the dependency info ones. The `pbxbuild::Phase::FrameworksResolver` would need to not hardcode the linker identifier to use — I'm not sure where to get the identifier to use, though...

grp:
This is good enough for now, though.

PR #168, opened 2016-10-20
Be safe when putting -mininum-deployment-target option

onhachoe:
if the option is not found and resolves to empty string,
it will have dangling -mininum-deployment-target option without value
and fail

### DIFF
--- a/Libraries/pbxbuild/Sources/Tool/InterfaceBuilderCommon.cpp
+++ b/Libraries/pbxbuild/Sources/Tool/InterfaceBuilderCommon.cpp
@@ -62,11 +62,11 @@ TargetedDeviceSetting(pbxsetting::Environment const &environment)
 std::vector<std::string> Tool::InterfaceBuilderCommon::
 DeploymentTargetArguments(pbxsetting::Environment const &environment)
 {
-    std::string min_deploy_target = environment.resolve(environment.resolve("DEPLOYMENT_TARGET_SETTING_NAME"));
-    if (min_deploy_target.empty()) {
-        return {};
+    std::string deploymentTarget = environment.resolve(environment.resolve("DEPLOYMENT_TARGET_SETTING_NAME"));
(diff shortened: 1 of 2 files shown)


grp:
When wouldn't there be a deployment target?

onhachoe:
DEPLOYMENT_TARGET_SETTING_NAME resolves to $(SDK)_DEPLOYMENT_TARGET, and if SDK's are careless and does not set this in its SDKSettings.plist, this fails.
Of course, its SDK's fault, but I think it doesn't hurt to have extra safe measure here.

Libraries/pbxbuild/Sources/Tool/InterfaceBuilderCommon.cpp
```diff
@@ -62,9 +62,11 @@ TargetedDeviceSetting(pbxsetting::Environment const &environment)
 std::vector<std::string> Tool::InterfaceBuilderCommon::
 DeploymentTargetArguments(pbxsetting::Environment const &environment)
 {
-    return {
-        "--minimum-deployment-target",
-        environment.resolve(environment.resolve("DEPLOYMENT_TARGET_SETTING_NAME")),
-    };
+    std::string min_deploy_target = environment.resolve(environment.resolve("DEPLOYMENT_TARGET_SETTING_NAME"));
+    if (min_deploy_target.empty()) {
+        return {};
```
onhachoe: Well, it feels to me kind of odd to be putting that in here.
It's a two level dereference we are making here on DEPLOYMENT_TARGET_SETTING_NAME,
and after all, we shouldn't be checking for sdk validity here.
Maybe we could add kind of validity checker at the end of SDK loads?

grp:
Travis is down, but good to retry when it's back.

grp:
Accidentally added a merge commit. Will need to force push to remove before landing.

PR #169, opened 2016-10-21
Pass down PATH environment varaible to invocation

onhachoe:
Tools invoked will sometimes expect PATH environment variable to be explicitly set
and contain the path the xcbuild searched for invocation.

### DIFF
--- a/Libraries/pbxbuild/Sources/Tool/ResolverCommon.cpp
+++ b/Libraries/pbxbuild/Sources/Tool/ResolverCommon.cpp
@@ -1,0 +1,75 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <pbxbuild/Tool/ResolverCommon.h>
+#include <sstream>
+
+namespace Tool = pbxbuild::Tool;
+
+static std::vector<std::string> sCommonPathEnvironmentExtensions = {
+    "$(PLATFORM_DEVELOPER_BIN_DIR)",
+    "$(DEVELOPER_BIN_DIR)",
+    "/usr/bin",
+    "/bin",
+    "/usr/sbin",
+    "/sbin",
+};
+
+static std::vector<std::string> sScriptPathEnvironmentExtensions = {
+    "$(DT_TOOLCHAIN_DIR)/usr/bin",
+    "$(DT_TOOLCHAIN_DIR)/usr/libexec",
+    "$(PLATFORM_DIR)/Developer/usr/bin",
+    "$(PLATFORM_DIR)/usr/local/bin",
+    "$(PLATFORM_DIR)/usr/bin",
+    "$(PLATFORM_DIR)/usr/local/bin",
+    "$(DEVELOPER_DIR)/usr/bin",
+    "$(DEVELOPER_DIR)/usr/local/bin",
+    "$(DEVELOPER_DIR)/Tools",
+    "/usr/bin",
+    "/bin",
+    "/usr/sbin",
+    "/sbin",
+};
+
+static std::string
+_PathEnvironmentVariable(
+    pbxsetting::Environment const &environment,
+    std::vector<std::string> &extensions)
+{
+    std::string pathEnvironmentVariable;
+    for (auto it = extensions.begin(); it != extensions.end(); it++) {
+        if (pathEnvironmentVariable.length()) {
+            pathEnvironmentVariable += ":";
+        }
+        pathEnvironmentVariable += environment.expand(pbxsetting::Value::Parse(*it));
+    }
+    return pathEnvironmentVariable;
+}
+
+std::unordered_map<std::string, std::string> Tool::ResolverCommon::
+commonExtendedEnvironmentVariables(
+    pbxsetting::Environment const &environment,
+    const std::unordered_map<std::string, std::string> &environmentVariables)
+{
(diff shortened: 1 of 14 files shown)


onhachoe:
Not really sure this is the way I should do it, requesting comments.

Libraries/xcexecution/Sources/Executor.cpp
```diff
@@ -23,3 +23,23 @@ Executor::
...
+{
+    if (environmentVariables.find("PATH") == environmentVariables.end()) {
+        // if PATH environment variable is not overriden,
+        // prepend current executablePaths
+        auto pathExtendedEnvironmentVariables = environmentVariables;
+        std::string pathEnvironmentVariable = "";
+        for (auto path : executablePaths) {
+            pathEnvironmentVariable += path + ":";
+        }
+        pathExtendedEnvironmentVariables["PATH"] = pathEnvironmentVariable + std::string(getenv("PATH"));
```
grp: Instead of `getenv()`, use `process::Context::environmentVariable()`. That allows simulating the environment for tests. (A `process::Context` is passed into this class.)

Libraries/xcexecution/Headers/xcexecution/Executor.h
```diff
@@ -55,6 +55,12 @@ class Executor {
         libutil::Filesystem *filesystem,
         pbxbuild::Build::Environment const &buildEnvironment,
         Parameters const &buildParameters) = 0;
+
+protected:
+    std::unordered_map<std::string, std::string> extendedEnvironmentVariables(
```
grp: This could be a static function in the `.cpp` file, I think — I don't see anything it needs from an instance.

Libraries/xcexecution/Sources/Executor.cpp
onhachoe: yeah, agreed

Libraries/xcexecution/Headers/xcexecution/Executor.h
onhachoe: agreed

onhachoe:
@grp putting it in where invocation object get initiated the first approached I've looked into, but that would entail replicate the similar logic into every resolvers we have, which looked bad.
Should we make base class for resolvers and put this in there?
From what I can see, having a base class would help a bit on gigantic if/else block on Libraries/pbxbuild/Sources/Phase/Context.cpp.

grp:
Yeah, I wanted to do a base class, but I wasn't sure how to handle the tools that need a different set of options. For example `Tool::CopyResolver` taking a log message title or some taking only one file and others taking many.

grp:
Ugh, please force push this branch to remove the merge commit I just added.

onhachoe:
changed to add environment variables from invocation creation stage.

Libraries/pbxbuild/Sources/Tool/ResolverCommon.cpp
```diff
@@ -0,0 +1,36 @@
...
+extendedEnvironmentVariables(
+    const std::vector<std::string> &executablePaths,
+    const std::unordered_map<std::string, std::string> &environmentVariables)
+{
+    if (environmentVariables.find("PATH") == environmentVariables.end()) {
+        std::string pathEnvironmentVariable;
+        for (auto path : executablePaths) {
+            pathEnvironmentVariable += path + ":";
+        }
+        auto processPathEnvironmentVariable = process::DefaultContext().environmentVariable("PATH");
```
grp: This is just as bad as `getenv()`: it's relying on external process state. To make it clear what state this function uses and to make it testable, it'd need a `process::Context const *` parameter passed in from above.

However, that would be a lot of work to thread through the layers into `pbxbuild::Tool`, and would be a bit unfortunate to require anyway. Luckily, there's an alternative: environment variables are populated in the build settings! So you can just get `PATH` from `environment.resolve("PATH")` on the `pbxsetting::Environment` already available.

onhachoe:
for tools needing PATH env variable, I've found some references
http://i.stack.imgur.com/4SryA.png
http://hongchaozhang.github.io/images/006_warning_type_from_build_log.png
https://developer.apple.com/library/content/technotes/tn2415/Art/tn2415_buildLog.png

grp:
An interesting thing to note about those screenshots: it isn't using `PATH` from the environment, but one that appears to be constructed as:

     $(PLATFORM_DEVELOPER_BIN_DIR):$(DEVELOPER_BIN_DIR):/usr/bin:/bin:/usr/sbin:/sbin

That makes sense, since it would prevent outside environment from changing the build.

grp:
And shell script build phases have a different `PATH`, including the toolchain:

    $(DT_TOOLCHAIN_DIR)/usr/bin
    $(DT_TOOLCHAIN_DIR)/usr/libexec
    $(PLATFORM_DIR)/Developer/usr/bin
    $(PLATFORM_DIR)/usr/local/bin
    $(PLATFORM_DIR)/usr/bin
    $(PLATFORM_DIR)/usr/local/bin
    $(DEVELOPER_DIR)/usr/bin
    $(DEVELOPER_DIR)/usr/local/bin
    $(DEVELOPER_DIR)/Tools
    /usr/bin
    /bin
    /usr/sbin
    /sbin

I'd assume that's equivalent to the executable search paths used for any tool, so scripts can easily execute tools in the same way.

onhachoe:
so then, what should I do?
just construct the path out of what we have in $(PLATFORM_DEVELOPER_BIN_DIR):$(DEVELOPER_BIN_DIR):/usr/bin:/bin:/usr/sbin:/sbin
and make exception for script phase to include what you have listed above?

grp:
Yeah, that seems reasonable to me: Hardcode the one for most tools to the format above, and make the one in `ScriptResolver` use the executable paths from `targetEnvironment.executablePaths()`.

grp:
In favor of #218.

PR #170, opened 2016-10-21
Add JSON support to property list utility. Fixes #160.

### LINKED ISSUE #160: Support JSON in plutil
It looks like the JSON parser and writer are already implemented. Not clear to me whether there's nontrivial work left or if it's just plugging the various bits and pieces into the pre-existing parsers and writers.

grp:


Libraries/plist/Tools/plutil.cpp
```diff
@@ -594,31 +640,42 @@ Modify(Filesystem *filesystem, Options const &options, std::string const &file,
...
+                    serialize = plist::Format::ASCII::Serialize(writeObject, ascii);
+                    break;
+                }
+                default: abort();
             }
-            case plist::Format::Type::ASCII: {
-                plist::Format::ASCII ascii = plist::Format::ASCII::Create(false, plist::Format::Encoding::UTF8);
-                out = plist::Format::Any::Create<plist::Format::ASCII>(ascii);
-                break;
+        } else if (ext::optional<Options::Format::Additional> additional = format->additional()) {
```
Ktwu: What's up with adding an "additional" format type?

grp: See the discussion on #165 — in short, `plist::Format::Type` doesn't include JSON (and I don't want to add it).

copumpkin:
w00t, I look forward to this going in 👍 👍

PR #173, opened 2016-10-23
Adjust loaded platform specifications for additional domains.

grp:
Prevents a warning about duplicate specifications on start.


grp:
Hmm, I'll do a force push to remove the merge commit without landing.

@sas is there a way to turn off requiring a merge before landing? Defeats the purpose of rebate and commit if you have to merge from master first.

grp:
without landing -> before landing

PR #174, opened 2016-10-27
Improve JSON parsing to handle more edge cases.

grp:
Using a JSON parser test suite, refine number and key parsing to match
the JSON specification. Remaining issues are related to Unicode support,
escape sequences, and trailing commas.

The last of those is intentionally non-conforming for compatibility.

### DIFF
--- a/Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
+++ b/Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
@@ -29,8 +29,7 @@
  *
  * Strings
  *
- *  JSON      '...'
- *            "..."  quoted chars like \', unicode chars, multiple-line
+ *  JSON      "..."  quoted chars like \", unicode chars, multiple-line
  *  ASCII:    '...'
  *            "..."  quoted chars like \', multiple-line
  *
@@ -216,7 +215,7 @@ ASCIIPListLexerReadData(ASCIIPListLexer *lexer)
 static int
 ASCIIPListLexerReadString(ASCIIPListLexer *lexer)
 {
-    if (lexer->pointer[0] == '\'') {
+    if (lexer->pointer[0] == '\'' && lexer->style != kASCIIPListLexerStyleJSON) {
         return ASCIIPListLexerReadSingleQuotedString(lexer);
     } else if (lexer->pointer[0] == '\"') {
         return ASCIIPListLexerReadDoubleQuotedString(lexer);
@@ -226,7 +225,7 @@ ASCIIPListLexerReadString(ASCIIPListLexer *lexer)
 }
 
 static int
-ASCIIPListLexerReadNumber(ASCIIPListLexer *lexer)
+ASCIIPListLexerReadJSONNumber(ASCIIPListLexer *lexer)
 {
     bool integer = true;
     char const *p = lexer->pointer;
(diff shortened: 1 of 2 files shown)


Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
```diff
@@ -216,7 +215,7 @@ ASCIIPListLexerReadData(ASCIIPListLexer *lexer)
 static int
 ASCIIPListLexerReadString(ASCIIPListLexer *lexer)
 {
-    if (lexer->pointer[0] == '\'') {
+    if (lexer->pointer[0] == '\'' && lexer->style == kASCIIPListLexerStyleASCII) {
         return ASCIIPListLexerReadSingleQuotedString(lexer);
     } else if (lexer->pointer[0] == '\"') {
```
Ktwu: Why doesn't the '\"' check get the style check too?

Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
```diff
@@ -226,27 +225,39 @@ ASCIIPListLexerReadString(ASCIIPListLexer *lexer)
...
         p++;
+    }
 
-    if (!isdigit(*p))
+    if (!isdigit(*p)) {
         return kASCIIPListLexerInvalidToken;
+    }
+    bool zero = (*p == '0');
+    while (isdigit(*p)) {
+        /* Numbers cannot start with zero. */
```
Ktwu: So something like "00000" is valid?

Ktwu: What about 0000.0000?

Libraries/plist/Tests/Format/test_JSON.cpp
```diff
@@ -98,3 +98,38 @@ TEST(JSON, Empty)
...
 
+TEST(JSON, SingleQuotes)
+{
+    /* Single quotes are not allowed in JSON. */
+    auto contents = Contents("{ 'one': 1 }");
+    auto deserialize = JSON::Deserialize(contents, JSON::Create());
+    EXPECT_EQ(deserialize.first, nullptr);
+}
+
+TEST(JSON, Number)
```
Ktwu: What about testing that the right number is decoded?

Ktwu: What about a test for -1?

Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
grp: Yeah, that is valid.

Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
grp: JSON keys are double-quoted:

    { "key" : "value" }

But not single-quoted:

     { 'key' : 'value' }

ASCII property lists support both.

Libraries/plist/Tests/Format/test_JSON.cpp
grp: good idea, will add -1. and test for the numbers

Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
grp: Yep, agree that's better.

Libraries/plist/Tests/Format/test_JSON.cpp
grp: Turns out I can't test the numbers because this is testing the cases where decoding fails so there's no number to check. :P

Libraries/plist/Sources/Format/ASCIIPListLexer.cpp
grp: Done.

PR #176, opened 2016-11-01
Fix and unit test for simple executor.

### LINKED ISSUE #175: CpHeader command failure
With the most recent commit (ef95e55f4732e26c83b4a8cb266f21c6897b309c) my build fails almost immediately in a `CpHeader` step with this log:

```
Create product structure
/bin/mkdir -p /Users/ksmiley/Library/Developer/Xcode/DerivedData/Lyft-dvyneupirgclsnggqtgohnrdrmxl/Build/Products/Debug-iphoneos/Alamofire-iOS/Alamofire.framework/Headers

CpHeader /Users/ksmiley/Library/Developer/Xcode/DerivedData/Lyft-dvyneupirgclsnggqtgohnrdrmxl/Build/Products/Debug-iphoneos/Alamofire-iOS/Alamofire.framework/Headers/Alamofire-iOS-umbrella.h '/Users/ksmiley/dev/phoenix/Pods/Alamofire/../Target Support Files/Alamofire-iOS/Alamofire-iOS-umbrella.h'
    cd /Users/ksmiley/dev/phoenix/Pods
    builtin-copy -exclude CVS -exclude .svn -exclude .git -exclude .hg /Users/ksmiley/dev/phoenix/Pods/Alamofire/../Target Support Files/Alamofire-iOS/Alamofire-iOS-umbrella.h /Users/ksmiley/Library/Developer/Xcode/DerivedData/Lyft-dvyneupirgclsnggqtgohnrdrmxl/Build/Products/Debug-iphoneos/Alamofire-iOS/Alamofire.framework/Headers

** BUILD FAILED **

The following build commands failed:
    CpHeader /Users/ksmiley/Library/Developer/Xcode/DerivedData/Lyft-dvyneupirgclsnggqtgohnrdrmxl/Build/Products/Debug-iphoneos/Alamofire-iOS/Alamofire.framework/Headers/Alamofire-iOS-umbrella.h '/Users/ksmiley/dev/phoenix/Pods/Alamofire/../Target Support Files/Alamofire-iOS/Alamofire-iOS-umbrella.h'
(1 failure)
```

I did a quick bisect and determined this is the first commit with this issue. Let me know if I can provide any more info about this!

grp:
Fixes #175 and starts looking into how to test highly integrated components like this.

### DIFF
--- a/Libraries/xcexecution/Sources/SimpleExecutor.cpp
+++ b/Libraries/xcexecution/Sources/SimpleExecutor.cpp
@@ -217,8 +217,7 @@ performInvocations(
     process::Context const *processContext,
     process::Launcher *processLauncher,
     Filesystem *filesystem,
-    pbxproj::PBX::Target::shared_ptr const &target,
-    pbxbuild::Target::Environment const &targetEnvironment,
+    std::vector<std::string> const &executablePaths,
     std::vector<pbxbuild::Tool::Invocation> const &orderedInvocations,
     bool createProductStructure)
 {
@@ -258,7 +257,8 @@ performInvocations(
                         processContext->groupID(),
                         processContext->userName(),
                         processContext->groupName());
-                    success = driver->run(&context, filesystem) != 0;
+                    int exitCode = driver->run(&context, filesystem);
+                    success = (exitCode == 0);
 
                     xcformatter::Formatter::Print(_formatter->finishInvocation(invocation, *builtin, createProductStructure));
                 } else {
@@ -273,7 +273,7 @@ performInvocations(
                         path = external;
                     }
                 } else {
-                    path = filesystem->findExecutable(*external, targetEnvironment.executablePaths());
+                    path = filesystem->findExecutable(*external, executablePaths);
                 }
 
                 if (path) {
@@ -329,13 +329,13 @@ buildTarget(
     }
 
     xcformatter::Formatter::Print(_formatter->beginCreateProductStructure(target));
-    std::pair<bool, std::vector<pbxbuild::Tool::Invocation>> structureResult = performInvocations(processContext, processLauncher, filesystem, target, targetEnvironment, *orderedInvocations, true);
+    std::pair<bool, std::vector<pbxbuild::Tool::Invocation>> structureResult = performInvocations(processContext, processLauncher, filesystem, targetEnvironment.executablePaths(), *orderedInvocations, true);
     xcformatter::Formatter::Print(_formatter->finishCreateProductStructure(target));
     if (!structureResult.first) {
         return structureResult;
     }
 
-    std::pair<bool, std::vector<pbxbuild::Tool::Invocation>> invocationsResult = performInvocations(processContext, processLauncher, filesystem, target, targetEnvironment, *orderedInvocations, false);
+    std::pair<bool, std::vector<pbxbuild::Tool::Invocation>> invocationsResult = performInvocations(processContext, processLauncher, filesystem, targetEnvironment.executablePaths(), *orderedInvocations, false);
     if (!invocationsResult.first) {
         return invocationsResult;
     }
--- a/Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
+++ b/Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
@@ -1,0 +1,145 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <gtest/gtest.h>
+#include <xcexecution/SimpleExecutor.h>
+#include <xcformatter/NullFormatter.h>
+#include <pbxbuild/Tool/Invocation.h>
+#include <builtin/Driver.h>
+#include <builtin/Registry.h>
+#include <process/MemoryContext.h>
+#include <process/MemoryLauncher.h>
+#include <libutil/MemoryFilesystem.h>
+
+using xcexecution::SimpleExecutor;
+using libutil::Filesystem;
+using libutil::MemoryFilesystem;
+
+class Driver : public builtin::Driver {
+public:
+    using Impl = std::function<int(process::Context const *processContext, Filesystem *filesystem)>;
+
+private:
+    std::string _name;
+    Impl        _impl;
+
+public:
+    Driver(std::string const &name, Impl const &impl) :
+        _name(name),
+        _impl(impl)
+    {
+    }
+
+public:
+    virtual std::string name()
+    { return _name; }
+
+public:
+    virtual int run(process::Context const *processContext, Filesystem *filesystem)
+    { return _impl(processContext, filesystem); }
+};
+
+TEST(SimpleExecutor, PropagateToolResult)
+{
+    /* Create in-memory execution environment. */
+    auto filesystem = MemoryFilesystem({
+        MemoryFilesystem::Entry::File("fail-tool", std::vector<uint8_t>()),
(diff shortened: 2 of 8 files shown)


Libraries/xcexecution/Sources/SimpleExecutor.cpp
```diff
@@ -258,7 +257,8 @@ performInvocations(
                         processContext->groupID(),
                         processContext->userName(),
                         processContext->groupName());
-                    success = driver->run(&context, filesystem) != 0;
+                    int exitCode = driver->run(&context, filesystem);
+                    success = (exitCode == 0);
```
grp: `==` instead of `!=` fixes the bug.

grp:
The unit test here has a lot of setup code. Ideas for how to shrink that down would be great!

grp:
Ping. This one is more important than usual since it's needed to fix any builds with builtin tools.

Libraries/xcformatter/Sources/NullFormatter.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <xcformatter/NullFormatter.h>
+#include <pbxbuild/Tool/Invocation.h>
+#include <pbxbuild/Build/Context.h>
+
+#include <algorithm>
+
+using xcformatter::NullFormatter;
+
+NullFormatter::
```
Ktwu: Do you really need to specify an empty constructor / destructor?

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
```diff
@@ -0,0 +1,145 @@
...
+        std::vector<std::string>(),
+        std::unordered_map<std::string, std::string>(),
+        0,
+        0,
+        "user",
+        "group");
+
+    /* Create test executor. */
+    auto formatter = xcformatter::NullFormatter::Create();
+    std::vector<std::string> const executablePaths = { "/" };
```
Ktwu: Sort of weird to define executablePaths in a code chunk that doesn't use it.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto registry = builtin::Registry::Create({
+        std::static_pointer_cast<builtin::Driver>(std::make_shared<Driver>("builtin-fail", [](process::Context const *context, Filesystem *filesystem) -> int {
+            return 1;
+        })),
+        std::static_pointer_cast<builtin::Driver>(std::make_shared<Driver>("builtin-success", [](process::Context const *context, Filesystem *filesystem) -> int {
+            return 0;
+        })),
+    });
+
+    auto context = process::MemoryContext(
```
Ktwu: Thoughts on defaults in the constructor?

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+TEST(SimpleExecutor, PropagateToolResult)
+{
+    /* Create in-memory execution environment. */
+    auto filesystem = MemoryFilesystem({
+        MemoryFilesystem::Entry::File("fail-tool", std::vector<uint8_t>()),
+        MemoryFilesystem::Entry::File("success-tool", std::vector<uint8_t>()),
+    });
+
+    auto launcher = process::MemoryLauncher({
```
Ktwu: Seems like the lowest hanging fruit for your test setup is to better unify the fake filesystem and the launcher.  Honestly, given the launcher's entries, you could create the filesystem.  Perhaps you can create a test helper function that iterates through a launcher's behavior and creates the filesystem underneath it?

Libraries/xcexecution/Sources/SimpleExecutor.cpp
Ktwu: Amazing.  Not really a fan of adding 'int exitCode' since it's used in this one case, but I don't care enough either way.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
```diff
@@ -0,0 +1,145 @@
...
+        { "/fail-tool", [](Filesystem *filesystem, process::Context const *context) -> ext::optional<int> {
+            return 1;
+        } },
+        { "/success-tool", [](Filesystem *filesystem, process::Context const *context) -> ext::optional<int> {
+            return 0;
+        } },
+    });
+
+    auto registry = builtin::Registry::Create({
```
Ktwu: Can you remind me what a builtin tool represents?  It'd be nice if this code could be slimmed down but I don't have any ideas.

Libraries/xcformatter/Sources/NullFormatter.cpp
grp: Probably not, I think this is because I copied it from the `DefaultFormatter` which has a constructor argument.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
grp: One of the tools starting with `builtin-` you see in builds, like `builtin-copy` or `builtin-infoPlistUtility`.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
grp: I'm wondering what they should be... anything actually *using* these values (especially the user/group ones) is going to fail, so it's relying on this specific code knowing that they aren't used... not sure that's a good thing to put into the API itself.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
grp: Ya, could move these three lines below the `Tool::Invocation` setup. None of them are used until after that, AFAICT.

Libraries/xcexecution/Sources/SimpleExecutor.cpp
grp: Made it a little more obvious that it's returning an `int` rather than a `bool`. I already missed the `!= 0` once when it was on the same line. 😶

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
grp: Yeah, I'm seeing this more and more in the filesystem too... for example, `isReadable()` is implicitly depending on the user and group (supposedly encapsulated in the the `process::Context`) to know whether the user/group/other permissions apply... and even `createFile()` implicitly uses the process-inherited `umask` for the permissions it sets.

If the process launcher and filesystem were merged, the memory version could treat the process body as the execute aspect of the filesystem entry, I guess! Would need to figure out the coupling that introduces. Even the `SimpleExecutor` runs built-in tools using an in-memory process context but the default filesystem.

Ktwu: Sort of out of scope for this patch, yeah? :)

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
Ktwu: Hm, yeah, probably not a good thing to use in the API then.

Ktwu:
Recommendation:  do whatever cleanup you'd like to do (like moving `executablePaths`), and I'll review again since this actually fixes a bug.

Cutting down on the unit test setup will be nice in the future, but unifying the filesystem+launcher will take some work (if it's even possible) and shouldn't hold this back.

Libraries/xcexecution/Tests/test_SimpleExecutor.cpp
grp: Probably 👍

Ktwu:
@sas, @onhachoe:  you two have any good ideas on how to trim down the unit test setup?

PR #177, opened 2016-11-04
made lsbom default usage print correct info

brtsai:


Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -455,7 +431,12 @@ main(int argc, char **argv)
             printf("%s\n", path.c_str());
         } else {
             // TODO: Respect options about what to print.
-            printf("%s\n", path.c_str());
+            std::stringstream octConvertingStream;
+            octConvertingStream << std::oct << ntohs(path_info_2_value->mode);
+            std::string mode = octConvertingStream.str();
```
grp: I would call this just `stream`, pretty clear it's being used for conversion from the `std::oct` in there.

brtsai: screen shot 2016-11-04 at 9 19 38 am

grp:
Update: there's still a few more properties that should be printed for files, in addition to these for directories.

PR #178, opened 2016-11-04
Filesystem functionality to replace /bin/cp, /bin/chmod, and chmod().

grp:
Moves a bunch of stuff into the filesystem and cleans up the API to be more consistent. Includes tests for filesystem changes.

### DIFF
--- a/Libraries/builtin/Sources/copy/Driver.cpp
+++ b/Libraries/builtin/Sources/copy/Driver.cpp
@@ -108,7 +108,7 @@ Run(Filesystem *filesystem, Options const &options, std::string const &workingDi
             input = filesystem->resolvePath(input);
         }
 
-        if (!filesystem->isDirectory(input) && !filesystem->isReadable(input)) {
+        if (filesystem->type(input) != Filesystem::Type::Directory && !filesystem->isReadable(input)) {
             if (options.ignoreMissingInputs()) {
                 continue;
             } else {
--- a/Libraries/builtin/Sources/copy/Driver.cpp
+++ b/Libraries/builtin/Sources/copy/Driver.cpp
@@ -20,6 +20,7 @@
 using builtin::copy::Driver;
 using builtin::copy::Options;
 using libutil::Filesystem;
+using libutil::Permissions;
 using libutil::FSUtil;
 
 Driver::
@@ -41,42 +42,54 @@ name()
 static bool
 CopyPath(Filesystem *filesystem, std::string const &inputPath, std::string const &outputPath)
 {
-    if (!filesystem->createDirectory(FSUtil::GetDirectoryName(outputPath), true)) {
+    /* Must be writable once copied to allow subsequent builds to overwrite. */
+    Permissions permissions;
+    permissions.user(Permissions::Permission::Write, true);
+    Permissions::Operation operation = Permissions::Operation::Add;
+
+    ext::optional<Filesystem::Type> type = filesystem->type(inputPath);
+    if (!type) {
+        /* Unknown type or doesn't exist. */
         return false;
     }
 
-    // TODO: This shouldn't use external tools to copy.
-    process::Context const *context = process::Context::GetDefaultUNSAFE();
-    process::Launcher *launcher = process::Launcher::GetDefaultUNSAFE();
+    switch (*type) {
+        case Filesystem::Type::File: {
+            if (!filesystem->copyFile(inputPath, outputPath)) {
+                return false;
+            }
 
-    process::MemoryContext cp = process::MemoryContext(
-        "/bin/cp",
-        context->currentDirectory(),
-        { "-R", inputPath, outputPath },
-        context->environmentVariables(),
-        context->userID(),
-        context->groupID(),
-        context->userName(),
-        context->groupName());
-    if (launcher->launch(filesystem, &cp).value_or(-1) != 0) {
-        return false;
+            if (!filesystem->writeFilePermissions(outputPath, operation, permissions)) {
+                return false;
+            }
+
+            return true;
+        }
+        case Filesystem::Type::SymbolicLink: {
+            if (!filesystem->copySymbolicLink(inputPath, outputPath)) {
+                return false;
+            }
+
+            if (!filesystem->writeSymbolicLinkPermissions(outputPath, operation, permissions)) {
+                return false;
+            }
+
+            return true;
+        }
+        case Filesystem::Type::Directory: {
+            if (!filesystem->copyDirectory(inputPath, outputPath, true)) {
+                return false;
+            }
+
+            if (!filesystem->writeDirectoryPermissions(outputPath, operation, permissions, true)) {
+                return false;
+            }
+
+            return true;
+        }
     }
 
-    /* Should preserve permissions but make writable. */
-    process::MemoryContext chmod = process::MemoryContext(
-        "/bin/chmod",
-        context->currentDirectory(),
-        { "-R", "+w", outputPath },
-        context->environmentVariables(),
-        context->userID(),
-        context->groupID(),
-        context->userName(),
-        context->groupName());
-    if (launcher->launch(filesystem, &chmod).value_or(-1) != 0) {
-        return false;
-    }
-
-    return true;
+    abort();
 }
 
 static int
@@ -99,7 +112,14 @@ Run(Filesystem *filesystem, Options const &options, std::string const &workingDi
     }
 
     std::string const &output = FSUtil::ResolveRelativePath(*options.output(), workingDirectory);
+    if (!filesystem->createDirectory(output, true)) {
+        return false;
+    }
+
+#if 0
+    // TODO: handle excludes paths
     auto excludes = std::unordered_set<std::string>(options.excludes().begin(), options.excludes().end());
+#endif
 
     for (std::string input : options.inputs()) {
         input = FSUtil::ResolveRelativePath(input, workingDirectory);
--- a/Libraries/builtin/Sources/copy/Driver.cpp
+++ b/Libraries/builtin/Sources/copy/Driver.cpp
@@ -41,7 +41,7 @@ name()
 static bool
 CopyPath(Filesystem *filesystem, std::string const &inputPath, std::string const &outputPath)
 {
-    if (!filesystem->createDirectory(FSUtil::GetDirectoryName(outputPath))) {
+    if (!filesystem->createDirectory(FSUtil::GetDirectoryName(outputPath), true)) {
         return false;
     }
 
--- a/Libraries/builtin/Tests/test_copy.cpp
+++ b/Libraries/builtin/Tests/test_copy.cpp
@@ -1,0 +1,115 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <gtest/gtest.h>
+#include <builtin/copy/Options.h>
+#include <builtin/copy/Driver.h>
+#include <libutil/Filesystem.h>
+#include <libutil/MemoryFilesystem.h>
+#include <process/Context.h>
+#include <process/MemoryContext.h>
+#include <plist/Format/Encoding.h>
+
+using builtin::copy::Driver;
+using builtin::copy::Options;
+using libutil::Filesystem;
+using libutil::MemoryFilesystem;
+
+static std::vector<uint8_t>
+Contents(std::string const &string)
+{
+    return std::vector<uint8_t>(string.begin(), string.end());
+}
+
+static process::MemoryContext
+Process(std::string const &executable, std::vector<std::string> const &arguments)
+{
+    return process::MemoryContext(
+        executable,
+        "/",
+        arguments,
+        std::unordered_map<std::string, std::string>(),
+        0,
+        0,
+        "root",
+        "wheel");
+}
+
+TEST(copy, Name)
+{
+    Driver driver;
+    EXPECT_EQ(driver.name(), "builtin-copy");
+}
+
+TEST(copy, CopyFiles)
+{
+    std::vector<uint8_t> contents;
+    MemoryFilesystem filesystem = MemoryFilesystem({
+        MemoryFilesystem::Entry::File("in1", Contents("one")),
+        MemoryFilesystem::Entry::File("in2", Contents("two")),
+        MemoryFilesystem::Entry::Directory("output", { }),
+    });
+
+    Driver driver;
+    auto process = Process(driver.name(), { "in1", "in2", "output", });
+    EXPECT_EQ(0, driver.run(&process, &filesystem));
+
+    contents.clear();
+    EXPECT_TRUE(filesystem.read(&contents, "/output/in1"));
+    EXPECT_EQ(contents, Contents("one"));
+
+    contents.clear();
+    EXPECT_TRUE(filesystem.read(&contents, "/output/in2"));
+    EXPECT_EQ(contents, Contents("two"));
+}
+
+TEST(copy, CopyDirectory)
+{
+    std::vector<uint8_t> contents;
+    MemoryFilesystem filesystem = MemoryFilesystem({
+        MemoryFilesystem::Entry::Directory("input", {
+            MemoryFilesystem::Entry::File("in1", Contents("one")),
+            MemoryFilesystem::Entry::File("in2", Contents("two")),
+        }),
+        MemoryFilesystem::Entry::Directory("output", { }),
+    });
+
+    Driver driver;
+    auto process = Process(driver.name(), { "input", "output", });
+    EXPECT_EQ(0, driver.run(&process, &filesystem));
+    EXPECT_EQ(filesystem.type("/output"), Filesystem::Type::Directory);
+
+    contents.clear();
+    EXPECT_TRUE(filesystem.read(&contents, "/output/input/in1"));
+    EXPECT_EQ(contents, Contents("one"));
+
+    contents.clear();
+    EXPECT_TRUE(filesystem.read(&contents, "/output/input/in2"));
+    EXPECT_EQ(contents, Contents("two"));
+}
+
+TEST(copy, IgnoreMissingOutput)
+{
+    std::vector<uint8_t> contents;
+    MemoryFilesystem filesystem = MemoryFilesystem({
+        MemoryFilesystem::Entry::File("exists", Contents("")),
+        MemoryFilesystem::Entry::Directory("output", { }),
+    });
+
+    Driver driver;
+
+    /* Missing input should fail if not allowed. */
+    auto fail = Process(driver.name(), { "exists", "missing", "output", });
+    EXPECT_NE(0, driver.run(&fail, &filesystem));
+
+    /* Missing input should succeed when allowed. */
+    auto succeed = Process(driver.name(), { "-ignore-missing-inputs", "exists", "missing", "output", });
+    EXPECT_EQ(0, driver.run(&succeed, &filesystem));
+}
+
--- a/Libraries/libutil/Sources/DefaultFilesystem.cpp
+++ b/Libraries/libutil/Sources/DefaultFilesystem.cpp
@@ -26,11 +26,32 @@
 #endif
 
 using libutil::DefaultFilesystem;
+using libutil::Filesystem;
 
 bool DefaultFilesystem::
 exists(std::string const &path) const
 {
     return ::access(path.c_str(), F_OK) == 0;
+}
+
+ext::optional<Filesystem::Type> DefaultFilesystem::
+type(std::string const &path) const
+{
+    struct stat st;
+    if (::lstat(path.c_str(), &st) < 0) {
+        return ext::nullopt;
+    }
+
+    if (S_ISREG(st.st_mode)) {
+        return Type::File;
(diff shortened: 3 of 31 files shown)


Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -178,6 +145,55 @@ write(std::vector<uint8_t> const &contents, std::string const &path)
     return true;
 }
 
+bool DefaultFilesystem::
+copyFile(std::string const &from, std::string const &to)
```
Ktwu: Several things:

1: Why block off some of the checks, like this->isFile, in the ifdef?  Seems valuable regardless.
2: copyFile() has a fallback implementation?  Why not write this in there?
3: It doesn't look like you're using anything particularly OS-specific.  What is Linux missing prevents you from using the logic in the #ifdef on it?

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -178,6 +145,55 @@ write(std::vector<uint8_t> const &contents, std::string const &path)
...
+#else
+    return Filesystem::copyFile(from, to);
+#endif
+}
+
+bool DefaultFilesystem::
+removeFile(std::string const &path)
+{
+    if (::unlink(path.c_str()) < 0) {
+        if (::unlink(path.c_str()) < 0) {
```
Ktwu: Why unlink twice?

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -178,6 +145,55 @@ write(std::vector<uint8_t> const &contents, std::string const &path)
...
+        }
+    }
+    return true;
+}
+
+bool DefaultFilesystem::
+isSymbolicLink(std::string const &path) const
+{
+    struct stat st;
+    if (::lstat(path.c_str(), &st) < 0) {
```
Ktwu: Style curiosity question:  any qualms to shortening this down with something like

  return (::lstat(path.c_str(), &st) == 0) && S_ISLNK(st.st_mode);

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -194,53 +210,197 @@ readSymbolicLink(std::string const &path) const
 bool DefaultFilesystem::
 writeSymbolicLink(std::string const &target, std::string const &path)
 {
-    if (symlink(target.c_str(), path.c_str())) {
+    if (::symlink(target.c_str(), path.c_str()) == 0) {
```
Ktwu: return ::symlink(...) != 0?

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -194,53 +210,197 @@ readSymbolicLink(std::string const &path) const
...
-        return realPath;
+        return S_ISDIR(st.st_mode);
+    }
+}
+
+bool DefaultFilesystem::
+createDirectory(std::string const &path, bool recursive)
+{
+    /* Get current mode. */
+    mode_t mask = ::umask(0);
```
Ktwu: would prefer a better name like currentMode.

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -194,53 +210,197 @@ readSymbolicLink(std::string const &path) const
...
+bool DefaultFilesystem::
+removeDirectory(std::string const &path, bool recursive)
+{
+    if (recursive) {
+        bool success = true;
+
+        this->readDirectory(path, recursive, [this, &path, &success](std::string const &name) {
+            std::string full = path + "/" + name;
+            if (this->isDirectory(full)) {
+                if (!this->removeDirectory(full, false)) {
```
Ktwu: Why false?  Shouldn't this be true?

Libraries/libutil/Sources/Filesystem.cpp
```diff
@@ -17,21 +17,100 @@ using libutil::Filesystem;
...
+    }
+
+    if (recursive) {
+        bool success = true;
+
+        success &= this->readDirectory(from, recursive, [this, &from, &to, &success](std::string const &path) {
+            std::string fromPath = from + "/" + path;
+            std::string toPath = to + "/" + path;
+
+            if (this->isSymbolicLink(fromPath)) {
```
Ktwu: ifelse chains make me nervous.  Any thoughts on representing the type of the file path as an enum (something like FileType type = identifyType(fromPath)) and using that?

Libraries/libutil/Sources/MemoryFilesystem.cpp
```diff
@@ -298,37 +264,140 @@ removeFile(std::string const &path)
...
+removeSymbolicLink(std::string const &path)
+{
+    return false;
+}
+
+bool MemoryFilesystem::
+isDirectory(std::string const &path) const
+{
+    return WalkPath<MemoryFilesystem::Entry const>(this, path, false, [](MemoryFilesystem::Entry const *parent, std::string const &name, MemoryFilesystem::Entry const *entry) -> MemoryFilesystem::Entry const * {
+        if (entry != nullptr && entry->type() != MemoryFilesystem::Entry::Type::Directory) {
```
Ktwu: Nit:  You could unify isDirectory() and isFile() with a helper function that takes a MemoryFilesystem::Entry::Type.

Libraries/libutil/Tests/test_MemoryFilesystem.cpp
```diff
@@ -251,21 +207,109 @@ TEST(MemoryFilesystem, Write)
...
+
+    EXPECT_TRUE(filesystem.exists("/file1"));
+    EXPECT_TRUE(filesystem.removeFile("/file1"));
+    EXPECT_FALSE(filesystem.exists("/file1"));
+
+    EXPECT_TRUE(filesystem.exists("/dir1/file2"));
+    EXPECT_TRUE(filesystem.removeFile("/dir1/file2"));
+    EXPECT_FALSE(filesystem.exists("/dir1/file2"));
+
+    EXPECT_FALSE(filesystem.removeFile("/dir1"));
```
Ktwu: Probably ought to verify that if removeFile reports failure, the filesystem isn't accidentally mutated.

Libraries/libutil/Tests/test_MemoryFilesystem.cpp
```diff
@@ -251,21 +207,109 @@ TEST(MemoryFilesystem, Write)
     EXPECT_FALSE(filesystem.exists("/invalid/new"));
 }
 
-TEST(MemoryFilesystem, ResolvePath)
+TEST(MemoryFilesystem, CopyFile)
```
Ktwu: tests #sogood

Libraries/libutil/Sources/DefaultFilesystem.cpp
grp: 👍

Libraries/libutil/Sources/DefaultFilesystem.cpp
grp: The outer `readDirectory()` is recursing, so no need to recurse again on the inside, it'll handle it.

Libraries/libutil/Sources/Filesystem.cpp
grp: Yeah, probably should. Would also cut the number of `stat()` calls in here by 2/3...

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -51,6 +52,168 @@ isExecutable(std::string const &path) const
...
+    struct stat st;
+    if (::lstat(path.c_str(), &st) < 0) {
+        return ext::nullopt;
+    }
+
+    return ModePermissions(st.st_mode);
+}
+
+ext::optional<Permissions> DefaultFilesystem::
+readDirectoryPermissions(std::string const &path) const
```
Ktwu: This looks exactly the same as readFilePermissions :(

Can you unify the API?  Or at least use a helper function?

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -51,6 +52,168 @@ isExecutable(std::string const &path) const
...
+    }
+
+    if (recursive) {
+        bool success = true;
+
+        success &= this->readDirectory(path, recursive, [this, &path, &operation, &permissions, &success](std::string const &name) {
+            std::string full = path + "/" + name;
+            if (this->isDirectory(full)) {
+                if (!this->writeDirectoryPermissions(full, operation, permissions, false)) {
+                    success = false;
```
Ktwu: you don't need to assign success as false, do you?  If you return 'false', and you're &ing the value to success, it'll get updated anyways.

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -51,6 +52,168 @@ isExecutable(std::string const &path) const
...
+
+    updated->combine(operation, permissions);
+    mode_t mode = PermissionsMode(*updated);
+
+    if (::chmod(path.c_str(), mode) != 0) {
+        return false;
+    }
+
+    if (recursive) {
+        bool success = true;
```
Ktwu: Move 'success' out of the block and just return 'success' at the bottom?  You could get rid of the 'if !success return false' chunk that way.

Libraries/libutil/Sources/MemoryFilesystem.cpp
```diff
@@ -158,6 +159,66 @@ isExecutable(std::string const &path) const
...
+        return ext::nullopt;
+    }
+
+    return AllPermissions();
+}
+
+ext::optional<Permissions> MemoryFilesystem::
+readSymbolicLinkPermissions(std::string const &path) const
+{
+    return ext::nullopt;
```
Ktwu: What's the reasoning behind saying "a symbolic link has no permissions" vs "a symbolic link has all permissions" for a memory filesystem?  I'm guessing that the former will mean that the code won't let you do anything useful with a symlink (which seems hard for unit tests to deal with).

Libraries/libutil/Sources/Permissions.cpp
```diff
@@ -0,0 +1,39 @@
...
+            return;
+        case Operation::Remove:
+            _flags &= ~other._flags;
+            _user &= ~other._user;
+            _group &= ~other._group;
+            _other &= ~other._other;
+            return;
+    }
+
+    abort();
```
Ktwu: Do you need abort()?  Won't the compiler fail if you don't exhaust all cases in the switch() statement?

Libraries/libutil/Sources/MemoryFilesystem.cpp
grp: Doesn't exist vs exists with all permissions. The memory filesystem doesn't support symlinks (yet)

Libraries/libutil/Sources/Permissions.cpp
grp: I've occasionally seen warnings about missing return after, not sure why but I guess because you could cast an invalid value to the enum being switched on?

Libraries/libutil/Sources/DefaultFilesystem.cpp
Ktwu: I see.  Ignore my comment then; it's full of ignorance.

Libraries/libutil/Sources/DefaultFilesystem.cpp
Ktwu: If it's old code, then OK I guess.  Mind checking what happens if you unlink only once?

Libraries/builtin/Sources/copy/Driver.cpp
```diff
@@ -41,42 +42,45 @@ name()
...
-    if (!filesystem->createDirectory(FSUtil::GetDirectoryName(outputPath), true)) {
-        return false;
-    }
+    /* Must be writable once copied to allow subsequent builds to overwrite. */
+    Permissions permissions;
+    permissions.user(Permissions::Permission::Write, true);
+    Permissions::Operation operation = Permissions::Operation::Add;
+
+    if (filesystem->isDirectory(inputPath)) {
+        if (!filesystem->copyDirectory(inputPath, outputPath, true)) {
```
Ktwu: return filesystem->copyDirectory() && filesystem->writeDirectoryPermissions()?

Same nit for symbolic links and files.

Libraries/builtin/Tests/test_copy.cpp
```diff
@@ -0,0 +1,115 @@
...
+TEST(copy, Name)
+{
+    Driver driver;
+    EXPECT_EQ(driver.name(), "builtin-copy");
+}
+
+TEST(copy, CopyFiles)
+{
+    std::vector<uint8_t> contents;
+    MemoryFilesystem filesystem = MemoryFilesystem({
```
Ktwu: Idea to cut down on unit test clutter:  'using namespace MemoryFilesystem::Entry' and update types as appropriate.

Libraries/xcexecution/Sources/SimpleExecutor.cpp
```diff
@@ -199,8 +200,15 @@ writeAuxiliaryFiles(
                 xcformatter::Formatter::Print(_formatter->setAuxiliaryExecutable(auxiliaryFile.path()));
 
                 if (!_dryRun) {
-                    // FIXME: This should use the filesystem.
-                    if (::chmod(auxiliaryFile.path().c_str(), S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH) != 0) {
+                    Permissions permissions;
```
Ktwu: This is obnoxiously verbose, and it's not the first blob like this in this PR :(

Any idea on an API to trim this?

grp: lol. yes it is pretty bad :(

Maybe could offer something like this:

    Permissions(
        Permissions::All,
        { Permissions::Read, Permissions::Execute },
        { Permissions::Read, Permissions::Execute })

With `std::initializer_list` or something parameters. Cuts it down to three lines versus up to 10?

Ktwu: I'd be OK with that.

grp:
New version switches to a type enum to switch on, also fixes most of your other comments.

This might cause bugs because it uses `lstat()` rather than `stat()` for everything, so symlinks are now always symlinks. Before, `isFile()` on a symlink could be true if the symlink pointed to a file, even though `isSymbolicLink()` was also true.

Libraries/libutil/Sources/DefaultFilesystem.cpp
grp: `readDirectory()` returns whether there was an error; the function passed to it returns whether to continue. So the inner function can return `false` and the outer return `true`, or the other way around.

Libraries/libutil/Sources/DefaultFilesystem.cpp
grp: I'm not quite sure what to do with these APIs. Ideally, there'd just be one read and write permissions API. But since (read+write) symlink permissions are non-portable and (write only) directory permissions need an extra `bool recursive` parameter, I'm not sure what's better than six different functions to be super clear. O_o

In terms of implementation, I think it'd save like one line to make a function and avoid copying the 4 lines inside? Not sure if the savings is worth the indirection...

Libraries/builtin/Sources/copy/Driver.cpp
grp: I like multiple lines because it makes it easier if some lines return more, like `ext::optional<std::string>` or something.

Ideally, all of these functions would return `Result<void>` or `Result<std::string>` as appropriate, then use something either like Rust's `try!` to skip the verbose error handling, but I don't think that's possible in C++. :(

Libraries/pbxspec/Sources/Manager.cpp
```diff
@@ -360,42 +360,53 @@ registerDomains(Filesystem const *filesystem, std::vector<std::pair<std::string,
...
-                context.defaultType = (file ? "FileType" : std::string());
+                    /* For *.pbfilespec files, default to FileType specifications. */
+                    bool file = FSUtil::GetFileExtension(path) == "pbfilespec";
+                    context.defaultType = (file ? "FileType" : std::string());
 
-                if (!filesystem->isDirectory(path)) {
+                    if (filesystem->type(path) != Filesystem::Type::Directory) {
 #if 0
-                    fprintf(stderr, "importing specification '%s'\n", path.c_str());
+                        fprintf(stderr, "importing specification '%s'\n", path.c_str());
```
Ktwu: Get rid of this?

Libraries/pbxspec/Sources/Manager.cpp
```diff
@@ -360,42 +360,53 @@ registerDomains(Filesystem const *filesystem, std::vector<std::pair<std::string,
...
-        } else if (filesystem->exists(domain.second)) {
+                    return true;
+                });
+                break;
+            }
+            case Filesystem::Type::SymbolicLink:
+            case Filesystem::Type::File: {
 #if 0
-            fprintf(stderr, "importing specification '%s'\n", domain.second.c_str());
+                fprintf(stderr, "importing specification '%s'\n", domain.second.c_str());
```
Ktwu: And this?

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -26,13 +26,34 @@
...
 
+ext::optional<Filesystem::Type> DefaultFilesystem::
+type(std::string const &path) const
+{
+    struct stat st;
+    if (::lstat(path.c_str(), &st) < 0) {
+        return ext::nullopt;
+    }
+
+    if (S_ISREG(st.st_mode)) {
```
Ktwu: Out of curiosity, would anything in xcbuild benefit from symbolic link knowledge being obscured?  Something like "I want to copy a file, I don't care if it's a real file or a symbolic link?"

Ktwu:
> This might cause bugs because it uses lstat() rather than stat() for everything, so symlinks are now always symlinks.

Interesting.  What *should* the correct behavior be?

grp:
Both to your top-level comment and the one on `S_ISREG()`:

Originally the file and directory methods here ignored symlinks, like most of the POSIX functions. Once there was code switching on type, though, it was hard to know what to do: should a `Filesystem::Type::SymbolicLink` get `copyFile()` or `copyDirectory()`? That points towards a single `copy()` function, which works but ends up with tons of special cases (read: magic) like Unix `cp`...

I think the best way would be for code accessing a file to be *required* to switch on the type in order to do *anything* with it. Make it easy to default to the usual behavior ignoring symlinks, but make it explicit.

I started looking into that API and discovered that POSIX filesystem APIs make it impossible to work with a filesystem with any kind of consistency guarantees. Even with extensions as implemented in Linux, macOS, etc, something as simple as deleting a file has an inherent race condition.

So for now I'll go with: it probably works mostly right, hopefully.

Ktwu:
> Make it easy to default to the usual behavior ignoring symlinks, but make it explicit.

Ok.  If I have a symlink, how would I distinguish whether the symlink is for a file or a directory?

> So for now I'll go with: it probably works right most of the time. Hopefully.

Can you at least build a few projects to make sure it won't immediately break the world?

grp:
> Ok. If I have a symlink, how would I distinguish whether the symlink is for a file or a directory?

Like this:

```c++
std::string path = "/some/item";
assert(filesystem->isSymbolicLink(path));

ext::optional<std::string> resolved = filesystem->readSymbolicLink(path);
assert(resolved != ext::nullopt);
std::string absolute = FSUtil::ResolveRelativePath(*resolved, FSUtil::GetDirectoryName(path));
    
bool file = filesystem->isFile(absolute);
bool directory = filesystem->isDirectory(absolute);
```

Not too simple. :( The middle 3 lines could instead be `std::string absolute = filesystem->resolvePath(path);`, but that's a legacy thing leftover from before I started refactoring.

I'd prefer something like this, in Swift:

```swift
/* Open a file descriptor to `/some`, so it can't be moved from underneath. */
guard let some: Handle = filesystem.open("/some") else {
    return false /* `/some` doesn't exist or can't be opened. */
}
guard case .Directory(let parent: Directory) = some else {
     return false /* `/some` isn't a directory. */
}

/* Open a file descriptor to `/some/item`, also can't be moved from underneath. */
guard let item: Handle = some.child("item") else {
    return false /* `/some/item` doesn't exist or can't be opened. */
}
guard case .SymbolicLink(let link: SymbolicLink) = item else {
    return false /* `/some/item` isn't a symlink. */
}

/* Resolve the symbolic link to a file descriptor to its target. */
guard let target: Handle = link.resolve() else {
    return false /* Symbolic link target doesn't exist. */
}
switch target {
    case .File(let file):
        handleFile(file) /* Target is a file. */
    case .Directory(let directory):
        handleDirectory(directory) /* Target is a directory. */
    case .SymbolicLink(let link):
        handleSymlink(link) /* Another symlink! */
        /* Might call for a `.resolve(recursive: Bool) -> Handle?` API to avoid
         * potentially infinite recursive symbolic links. Or a convenience API like
         * `.resolveFile() -> File?` that recursively resolves only to a file. */
}
```

One issue with this idea is that only BSD supports `open(2)` on symlinks at all. And without opening it, there's no way to guarantee the path won't be swapped out for something else between calls.

> Can you at least build a few projects to make sure it won't immediately break the world?

Did that, although all of the project and spec loading code calls `realpath()` before opening anything, so symlinks are resolved anyway.

Ktwu:
Thanks for answering my questions.  I'm OK with this as-is.

PR #179, opened 2016-11-04
Partially update asset catalog compiler specification and resolver.

grp:
Asset catalog compilation now uses a custom variable to pass inputs
to the specification, and a custom grouping method to group inputs.

Does not pass sticker pack translations to asset catalog compiler,
as this seems to require parsing the asset catalog just to determine
which strings files should be used. Left as a future task.

### DIFF
--- a/Libraries/pbxbuild/Sources/Tool/AssetCatalogResolver.cpp
+++ b/Libraries/pbxbuild/Sources/Tool/AssetCatalogResolver.cpp
@@ -13,8 +13,11 @@
 #include <pbxbuild/Tool/OptionsResult.h>
 #include <pbxbuild/Tool/Tokens.h>
 #include <pbxbuild/Tool/Context.h>
+#include <pbxsetting/Type.h>
+#include <libutil/FSUtil.h>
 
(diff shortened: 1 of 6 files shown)


Libraries/pbxbuild/Sources/Tool/AssetCatalogResolver.cpp
```diff
@@ -28,19 +31,41 @@ resolve(
     pbxsetting::Environment const &baseEnvironment,
     std::vector<Phase::File> const &inputs) const
 {
+    std::vector<std::string> absoluteInputPaths;
+    std::vector<std::string> assetPaths;
+    std::vector<std::string> stickerPackStrings;
+    for (Phase::File const &input : inputs) {
+        if (input.fileType()->identifier() == "text.plist.strings") {
+            std::string strings;
```
sas: Nit: you could allocate your `strings` variable and then `reserve()` the amount of space you're going to need before doing the concatenations. That'll avoid a bunch of reallocations. This might not even be a hot path, so feel free to ignore the suggestion.

PR #180, opened 2016-11-04
fix bug in SimpleExecutor builtin tool execution

onhachoe:
builtin tools return 0 on success.


grp:
Thanks, also fixed in #176.

PR #181, opened 2016-11-04
handle GCC_PREPROCESSOR_DEFINITIONS

onhachoe:


Libraries/pbxbuild/Sources/Tool/ClangResolver.cpp
```diff
@@ -242,6 +249,7 @@ resolveSource(
     Tool::CompilerCommon::AppendIncludePathFlags(&arguments, env, toolContext->searchPaths(), headermapInfo);
     AppendFrameworkPathFlags(&arguments, env, toolContext->searchPaths());
     AppendCustomFlags(&arguments, env, fileType->GCCDialectName());
+    AppendPreprocessorDefinitionFlags(&arguments, env);
```
grp: Could you change `SwiftResolver` to use this function as well rather than duplicating the logic?

grp:
Took a look and I think #182 is how this should work. My bad on getting that wrong in the specification.

PR #183, opened 2016-11-05
Support reading JSON (as well as writing) in property list utility.

grp:
Also, fix default output format when performing modifications without
conversion. It would always fail due to not having a target format.


copumpkin:
👍

copumpkin:
🍾

PR #184, opened 2016-11-06
Support short options in lsbom.

grp:
Short options can be combined and used as a single flag. For split
short options, the value can follow in a combined option or in the
next option.

### DIFF
--- a/Libraries/libbom/Tools/lsbom.cpp
+++ b/Libraries/libbom/Tools/lsbom.cpp
@@ -114,67 +114,89 @@ Options::
 {
 }
 
+static ext::optional<std::vector<Options::PrintItem>>
+ParsePrintFormat(std::string const &print)
+{
+    std::vector<Options::PrintItem> format;
+
+    for (char c : print) {
+        switch (c) {
+            case 'c': format.push_back(Options::PrintItem::Checksum); break;
+            case 'f': format.push_back(Options::PrintItem::FileName); break;
+            case 'F': format.push_back(Options::PrintItem::FileNameQuoted); break;
+            case 'g': format.push_back(Options::PrintItem::GroupID); break;
+            case 'G': format.push_back(Options::PrintItem::GroupName); break;
+            case 'm': format.push_back(Options::PrintItem::Permissions); break;
+            case 'M': format.push_back(Options::PrintItem::PermissionsText); break;
+            case 's': format.push_back(Options::PrintItem::FileSize); break;
+            case 'S': format.push_back(Options::PrintItem::FileSizeFormatted); break;
+            case 't': format.push_back(Options::PrintItem::ModificationTime); break;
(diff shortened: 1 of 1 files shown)


Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -114,67 +114,91 @@ Options::
 {
 }
 
+static ext::optional<std::vector<Options::PrintItem>>
+ParsePrintFormat(std::string const &print)
+{
+    auto format = std::vector<Options::PrintItem>();
+
+    for (char c : print) {
+        switch (c) {
```
Ktwu: Hmm...is it possible you do do something more clever, like create a map for characters -> options, then just reference that?

grp: Yah, although it'd be a few more lines of code and a little slower. :( `std::unordered_map` requires allocation and can't be `constexpr`, so it'd have to allocate memory at least once. Would also take an extra line to search in the map than just print out. A better language would make this easy...

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -114,67 +114,91 @@ Options::
...
-        if (!result.first) {
-            return result;
-        }
+    if (!arg.empty() && arg[0] == '-') {
+        if (arg == "--help") {
+            return libutil::Options::Current<bool>(&_help, arg);
+        } else if (arg == "--arch") {
+            return libutil::Options::Next<std::string>(&_arch, args, it);
+        } else {
+            std::pair<bool, std::string> result = std::make_pair(true, std::string());
```
sas: Here you can use `auto result = std::make_pair(true, std::string());`.

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -114,67 +114,91 @@ Options::
 {
 }
 
+static ext::optional<std::vector<Options::PrintItem>>
+ParsePrintFormat(std::string const &print)
+{
+    auto format = std::vector<Options::PrintItem>();
```
grp: 👍

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -114,67 +114,91 @@ Options::
...
-        return libutil::Options::Current<bool>(&_noModes, arg, it);
-    } else if (arg == "--arch") {
-        return libutil::Options::Next<std::string>(&_arch, args, it);
-    } else if (arg == "-p") {
-        ext::optional<std::string> print;
-        auto result = libutil::Options::Next<std::string>(&print, args, it);
-        if (!result.first) {
-            return result;
-        }
+    if (!arg.empty() && arg[0] == '-') {
```
grp: 👍

PR #190, opened 2016-11-09
fix builtin pbxcp

onhachoe:
pbx cp command should specify output dir, not the output file path.

### DIFF
--- a/Specifications/Tool/com.apple.compilers.pbxcp.xcspec
+++ b/Specifications/Tool/com.apple.compilers.pbxcp.xcspec
@@ -11,7 +11,7 @@
     Type = Tool;
     Identifier = com.apple.compilers.pbxcp;
     Name = "Copy Utility";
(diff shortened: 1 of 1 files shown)


Specifications/Tool/com.apple.compilers.pbxcp.xcspec
```diff
@@ -11,7 +11,7 @@
     Type = Tool;
     Identifier = com.apple.compilers.pbxcp;
     Name = "Copy Utility";
-    CommandLine = "builtin-copy [options] [inputs] [output]";
+    CommandLine = "builtin-copy [options] [inputs] $(OutputPath:dir)";
```
grp: I think this was wrong, but in a different way: `CopyResolver.cpp` already adds the output directory to the end. This can just have the options:

     ExecPath = "builtin-copy";

onhachoe:
@grp  yeah, you are right! making change

PR #192, opened 2016-11-12
Start cleaning up tool resolver implementations.

grp:


Libraries/pbxbuild/Headers/pbxbuild/Tool/Input.h
```diff
@@ -29,51 +29,63 @@ class Environment;
...
+     * The resolved absolute path to the file.
      */
-    Target::BuildRules::BuildRule::shared_ptr const &buildRule() const
-    { return _buildRule; }
+    std::string const &path() const
+    { return _path; }
 
     /*
-     * The type of this file.
+     * The type of this file. Optional.
```
Ktwu: Optional?  Then why doesn't this return an optional type? :)

Libraries/pbxbuild/Headers/pbxbuild/Tool/Input.h
```diff
@@ -29,51 +29,63 @@ class Environment;
...
-     * The resolved absolute path to the file.
+     * The build rule saying *how* to build this file. Optional.
      */
-    std::string const &path() const
-    { return _path; }
+    Target::BuildRules::BuildRule::shared_ptr const &buildRule() const
+    { return _buildRule; }
+
+    /*
+     * A disambiguation identifier for files that have multiple with the
```
Ktwu: "that have multiple with the same base name"

Confusing.  You probably ought to reword this to something like "A disambiguation identifier for different files with the same base name..."

Libraries/pbxbuild/Headers/pbxbuild/Tool/Input.h
```diff
@@ -29,51 +29,63 @@ class Environment;
  */
```
Ktwu: Remove reference to project objects from tool input: LGTM

Libraries/pbxbuild/Sources/Phase/Context.cpp
```diff
@@ -263,7 +263,7 @@ Group(std::vector<Tool::Input> const &files)
          * the ungrouped  inputs so they don't get added again to the result.
          */
         ungrouped.erase(std::remove_if(ungrouped.begin(), ungrouped.end(), [&](Tool::Input const &ungroupedFile) {
-            if (ungroupedFile.fileType()->identifier() == "text.plist.strings") {
+            if (ungroupedFile.fileType() != nullptr && ungroupedFile.fileType()->identifier() == "text.plist.strings") {
```
Ktwu: Once again, this seems like it ought to be an optional.

Libraries/pbxbuild/Sources/Phase/Context.cpp
```diff
@@ -323,24 +324,28 @@ resolveBuildFiles(
 
             if (buildRule != nullptr) {
                 if (pbxspec::PBX::Tool::shared_ptr const &tool = buildRule->tool()) {
-                    // Some tools additionally limit their file types beyond what their build rule allows.
-                    // For example, the default compiler limits itself to just source files, despite its
-                    // default build rule specifying that it accepts all C-family inputs, including headers.
+                    /*
+                     * Some tools additionally limit their file types beyond what their build rule allows.
+                     * For example, the default compiler limits itself to just source files, despite its
+                     * default build rule specifying that it accepts all C-family inputs, including headers.
```
Ktwu: Why is this burden on the tool?  Why wouldn't the default compiler's default build rule exclude headers?

Libraries/pbxbuild/Sources/Phase/FrameworksResolver.cpp
```diff
@@ -84,21 +84,23 @@ resolve(Phase::Environment const &phaseEnvironment, Phase::Context *phaseContext
...
             if (it != phaseContext->toolContext().variantArchitectureInvocations().end()) {
                 std::vector<Tool::Invocation> const &sourceInvocations = it->second;
                 for (Tool::Invocation const &invocation : sourceInvocations) {
                     for (std::string const &output : invocation.outputs()) {
                         // TODO(grp): Is this the right set of source outputs to link?
+                        // TODO(grp): Use the object file file type and include in input.
                         if (FSUtil::GetFileExtension(output) == "o") {
-                            sourceOutputs.push_back(output);
+                            Tool::Input outputInput = Tool::Input(output, nullptr);
+                            sourceOutputs.push_back(outputInput);
```
Ktwu: sourceOutputs.push_Back(Tool::Input(output, nullptr));

Libraries/pbxbuild/Headers/pbxbuild/Tool/CopyResolver.h
```diff
@@ -34,13 +34,6 @@ class CopyResolver {
         std::string const &outputDirectory,
```
Ktwu: Always pass in inputs as tool inputs, rather than path strings: LGTM since I commented with nits.

Libraries/pbxbuild/Headers/pbxbuild/Phase/Context.h
```diff
@@ -87,14 +87,14 @@ class Context {
     /*
      * Groups the files according to the tools used to build them.
      */
-    static std::vector<std::vector<Phase::File>> Group(std::vector<Phase::File> const &files);
+    static std::vector<std::vector<Tool::Input>> Group(std::vector<Tool::Input> const &files);
```
grp: Yeah, they're a wrapper around a file path with extra info.

Libraries/pbxbuild/Sources/Phase/Context.cpp
grp: `pbxspec::PBX::Specification` has to be a pointer since it's `virtual`. And `optional<SomePointer>` is confusing because it could be `nullopt`, `nullptr`, or a value.

Two options: make a non-nullable pointer type and wrap that in `optional`, or check `nullptr` like this does. Neither are great. (Rust gets this right, too...)

Libraries/pbxbuild/Sources/Phase/Context.cpp
grp: The issue is that it takes `sourcecode.cpp`, and `sourcecode.cpp.header` inherits from `sourcecode.cpp`. I tried not checking the inheritance hierarchy, but that failed for `text.plist.xml` inherits from `text.plist` and needs to be an input to the plist copy tool...

Libraries/pbxbuild/Headers/pbxbuild/Tool/Input.h
grp: 👍

Libraries/pbxbuild/Headers/pbxbuild/Tool/AssetCatalogResolver.h
```diff
@@ -10,14 +10,20 @@
 #ifndef __pbxbuild_Tool_AssetCatalogResolver_h
```
Ktwu: Start cleaning up tool resolver implementations: LGTM

Libraries/pbxbuild/Headers/pbxbuild/Tool/Context.h
```diff
@@ -29,24 +29,27 @@ namespace Tool {
 
 class Context {
 private:
-    xcsdk::SDK::Target::shared_ptr _sdk;
+    xcsdk::SDK::Target::shared_ptr   _sdk;
     std::vector<xcsdk::SDK::Toolchain::shared_ptr> _toolchains;
```
Ktwu: Ugggh the spacing here for _toolchains is unsatisfying.

Libraries/pbxbuild/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ add_library(pbxbuild SHARED
             Sources/DerivedDataHash.cpp
```
Ktwu: > Auxiliary files were attached to invocations but had no relation to
the containing invocation.

I see, so auxiliary files are basically used by Ninja, not the tools that Ninja invokes.

I actually thought it was nice that invocations contained their auxiliary files, since in a way such files a dependency for the invocation.

Oh well.  LGTM

Libraries/pbxbuild/Headers/pbxbuild/Phase/File.h
```diff
@@ -29,73 +26,16 @@ class Environment;
...
-    /*
-     * Resolves build files into `Phase::File`s, using the phase environment. Note that due to
-     * certain types of build files being more complex than a single `Phase::File`, the number
-     * of `Phase::File`s returned may be different than the number of build files passed in.
+     * Resolves build files into `Tool::Input`s, using the phase environment. Note that due to
+     * certain types of build files being more complex than a single `Tool::Input`, the number
+     * of `Tool::Input`s returned may be different than the number of build files passed in.
      */
-    static std::vector<Phase::File>
+    static std::vector<Tool::Input>
```
Ktwu: Ok.

Libraries/pbxbuild/Sources/Phase/Context.cpp
Ktwu: > sourcecode.cpp.header inherits from sourcecode.cpp
:( ok

Libraries/pbxbuild/CMakeLists.txt
grp: Yeah, that was the original idea. I'd tried using that to put them in the dependency graph, but projects expect all of them to be written first and then used in custom scripts, etc. So they're already all being written before any invocations run.

In a some cases (scripts, linker file lists, Swift input) they were attached to the invocations that needed them. In a few cases (PCH info) they were totally independent. In others (header maps, module maps) they were dependencies but attached to the wrong invocation because the right one wasn't available.

Fixing the last set to work as dependencies was going to be pretty complicated especially for ones like module maps that are shared between multiple invocations, and not too useful due to the script input issue, so I sorta gave up on the idea.

Ktwu:
I don't think I had any major recommendations so overall LGTM.

PR #193, opened 2016-11-12
Complete newly added specifications.

grp:
Note Yacc and Lex are missing tool resolvers still, so won't work even with the specifications.


Specifications/Compiler/CMakeLists.txt
```diff
@@ -13,10 +13,8 @@ install(FILES
         com.apple.build-tasks.copy-png-file.xcspec
         com.apple.build-tasks.copy-strings-file.xcspec
         com.apple.build-tasks.ls-register-url.xcspec
-        com.apple.build-tools.nmedit.xcspec
         com.apple.compilers.assetcatalog.xcspec
         com.apple.compilers.gcc.xcspec
-        com.apple.compilers.lex.xcspec
```
sas: Maybe a dumb question: why do you remove these entries from this list?

grp: Moved them under Tool since they're Tool and not Compiler

matthewbauer:
@grp Okay, so can't we just make a generic ToolResolver to support any "tool", or do you think it's best to have one file for each tool?

grp:
@matthewbauer There is a `ToolResolver` that works for most tools, but Lex and Yacc need some extra logic. For example `LEXFLAGS` and `YACCFLAGS` need to be appended in code, `YACC_GENERATED_FILE_STEM` checked & used, etc.

Issue #195, opened 2016-11-16
xcbuild: infinite loop when inheriting specifications


Ktwu:
On Linux, if I run xcbuild without any input, I can get a crash that looks something like:

(gdb) bt
#0  0x00007ffff7d7013c in pbxspec::Manager::inheritSpecification(std::shared_ptr<pbxspec::PBX::Specification> const&) (this=0x402af0, specification=..., this=<optimized out>)
    at ../xcbuild/Libraries/pbxspec/Sources/Manager.cpp:332
#1  0x00007ffff7d73b9f in pbxspec::Manager::registerDomains(libutil::Filesystem const*, std::vector<std::pair<std::string, std::string>, std::allocator<std::pair<std::string, std::string> > > const&) (this=0x0, filesystem=
    0x7fffffffdcd0, domains=...) at ../xcbuild/Libraries/pbxspec/Sources/Manager.cpp:435
#2  0x00007ffff7f8d553 in pbxbuild::Build::Environment::Default(process::Context const*, libutil::Filesystem const*) (processContext=processContext@entry=0x7fffffffdce0, filesystem=filesystem@entry=0x7fffffffdcd0)
    at ../xcbuild/Libraries/pbxbuild/Sources/Build/Environment.cpp:73
#3  0x00007ffff7fe78ab in xcdriver::BuildAction::Run(process::Context const*, process::Launcher*, libutil::Filesystem*, xcdriver::Options const&) (processContext=processContext@entry=0x7fffffffdce0, processLauncher=processLauncher@entry=0x7fffffffdcf0, filesystem=filesystem@entry=0x7fffffffdcd0, options=...) at ../xcbuild/Libraries/xcdriver/Sources/BuildAction.cpp:152
#4  0x00007ffff7fe4e58 in xcdriver::Driver::Run(process::Context const*, process::Launcher*, libutil::Filesystem*) (processContext=processContext@entry=0x7fffffffdce0, processLauncher=processLauncher@entry=0x7fffffffdcf0, filesystem=filesystem@entry=0x7fffffffdcd0) at ../xcbuild/Libraries/xcdriver/Sources/Driver.cpp:56
#5  0x0000000000400c43 in main(int, char**) (argc=<optimized out>, argv=<optimized out>) at ../xcbuild/Libraries/xcdriver/Tools/xcbuild.cpp:23

When I enable debug logging, I see repeated ad-nauseum:

[...]

PR #196, opened 2016-11-16
Support loading specifications from a directory symlink.

grp:
@matthewbauer This should fix your issue — let me know if it doesn't. @Ktwu, curious if this helps with #195 at all.


matthewbauer:
This fixes this issue but now I think I'm running into #195.

PR #197, opened 2016-11-17
Fix a typo (copy-paste error?) in xcrun

sas:


Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -302,7 +302,7 @@ static int Run(Filesystem *filesystem, process::Context const *processContext, p
              return -1;
         }
     } else {
-        target = target = manager->findTarget(defaultSDK);
+        target = manager->findTarget(defaultSDK);
```
Ktwu: Ooops.

PR #200, opened 2016-11-18
Fix creating directories recursively.

grp:
cc @matthewbauer


grp:
Abandoning for #199 which is identical.

matthewbauer:
🐎

PR #201, opened 2016-11-18
Various patches to all parts of xcbuild.

grp:
Various patches to all parts of xcbuild. Highlights:

 - Cleaned up binary plist parser. Ended up -250 lines and much more readable.
 - Cleaned up specification types.
 - Fixed a bunch of warnings detected by MSVC. Mostly around casting.
 - Added conditional checks around various non-portable language extensions.
 - Various other fixes.


grp:
In favor of #202.

PR #202, opened 2016-11-18
Various patches to all parts of xcbuild.

grp:
Various patches to all parts of xcbuild. Highlights:

Cleaned up binary plist parser. Ended up -250 lines and much more readable.
Cleaned up specification types.
Fixed a bunch of warnings detected by MSVC. Mostly around casting.
Added conditional checks around various non-portable language extensions.
Various other fixes.


Libraries/pbxspec/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@
 
 add_library(pbxspec SHARED
```
Ktwu: "Use type rather than character contents for specification types": LGTM

Libraries/pbxbuild/Sources/Phase/File.cpp
```diff
@@ -38,7 +38,7 @@ ResolveBuildFiles(Filesystem const *filesystem, Phase::Environment const &phaseE
             continue;
         }
 
-        std::string fileNameDisambiguator;
+        ext::optional<std::string> fileNameDisambiguator;
```
Ktwu: Avoid always setting file disambiguation identifier: LGTM

I'm confused how you're getting away with changing a type here without things breaking.

Libraries/builtin/Sources/copy/Options.cpp
```diff
@@ -27,17 +27,17 @@ parseArgument(std::vector<std::string> const &args, std::vector<std::string>::co
     std::string const &arg = **it;
```
Ktwu: Avoid incorrect third parameter for command-line option parsing: LGTM

Libraries/ext/Headers/ext/optional
```diff
@@ -168,6 +168,14 @@ namespace std { namespace experimental { inline namespace fundamentals_v1 {
 #endif
```
Ktwu: Make compiler attributes more portable: @sas should review this commit.

Libraries/graphics/Sources/Format/PNG.cpp
```diff
@@ -480,7 +480,7 @@ Write(Image const &image)
     strm.next_in = buffer.data();
 
     /* Maximum compressed size is x1.01 + 12 the uncompressed size. */
-    std::vector<uint8_t> compressed = std::vector<uint8_t>(buffer.size() * 1.01 + 12);
+    std::vector<uint8_t> compressed = std::vector<uint8_t>(static_cast<size_t>(buffer.size() * 1.01) + 12);
```
Ktwu: Does/should this round up or down?

Libraries/libbom/Sources/bom.c
```diff
@@ -334,11 +334,11 @@ bom_variable_iterate(struct bom_context *context, bom_variable_iterator iterator
     context->iteration_count++;
 
     struct bom_header *header = (struct bom_header *)context->memory.data;
-    struct bom_variables *vars = (struct bom_variables *)((void *)header + ntohl(header->variables_offset));
+    struct bom_variables *vars = (struct bom_variables *)((uintptr_t)header + ntohl(header->variables_offset));
 
     ptrdiff_t var_offset = 0;
     for (size_t i = 0; i < ntohl(vars->count); i++) {
-        struct bom_variable *var = (struct bom_variable *)((void *)vars->first + var_offset);
+        struct bom_variable *var = (struct bom_variable *)((uintptr_t)vars + sizeof(struct bom_variable) + var_offset);
```
Ktwu: What happened here?  Why the change to the expression alongside the casting changes?

Libraries/libbom/Sources/bom.c
```diff
@@ -395,12 +395,12 @@ bom_variable_add(struct bom_context *context, const char *name, int data_index)
...
 
     struct bom_header *header = (struct bom_header *)context->memory.data;
-    struct bom_variables *vars = (struct bom_variables *)((void *)header + ntohl(header->variables_offset));
+    struct bom_variables *vars = (struct bom_variables *)((uintptr_t)header + ntohl(header->variables_offset));
 
     /* Find the end of the variables section. */
     ptrdiff_t var_offset = 0;
     for (size_t i = 0; i < ntohl(vars->count); i++) {
-        struct bom_variable *var = (struct bom_variable *)((void *)vars->first + var_offset);
+        struct bom_variable *var = (struct bom_variable *)((uintptr_t)vars + sizeof(struct bom_variable) + var_offset);
```
Ktwu: Same question here re: changing the expression.

Libraries/builtin/Sources/infoPlistUtility/Driver.cpp
```diff
@@ -126,7 +126,7 @@ AddBuildEnvironment(plist::Dictionary *root, pbxsetting::Environment const &envi
             std::string entry = (noff == std::string::npos ? targetedDeviceFamily.substr(off) : targetedDeviceFamily.substr(off, noff));
```
Ktwu: Correct casts and pointer arithmetic: Overall LGTM

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -239,7 +239,10 @@ Load(
         }
```
Ktwu: Fix rendition serialization to be operating system independent: LGTM

Libraries/xcsdk/Sources/SDK/Manager.cpp
```diff
@@ -47,7 +47,9 @@ findTarget(std::string const &name) const
 
         /* If the platform name matches but no targets do, use any target. */
         if (platform->name() == name || platform->path() == name) {
-            return platform->targets().back();
+            if (!platform->targets().empty()) {
+                return platform->targets().back();
```
Ktwu: Honestly it seems like an error case to have a platform with no SDKs.

Avoid returning a default SDK for a platform with no SDKs: LGTM

Libraries/xcexecution/Sources/SimpleExecutor.cpp
```diff
@@ -78,7 +78,7 @@ build(
         if (!targetEnvironment) {
```
Ktwu: Don't continue the build if target environment creation fails: LGTM

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -172,45 +172,20 @@ CommandHelp()
...
     std::string::size_type prev = 0;
     do {
         std::string::size_type pos = keyPathString.find(':', prev);
         std::string::size_type end = (pos == std::string::npos ? pos : pos - prev);
+        bool emptyInitialKey = (prev == 0 && pos == 0);
 
-        std::string key = keyPathString.substr(prev, end);
-        if (!key.empty()) {
-            keyPath->push(key);
+        if (!emptyInitialKey) {
```
Ktwu: Idiot question: what would an empty initial key mean for navigating the plist?

Libraries/plist/CMakeLists.txt
```diff
@@ -66,11 +66,11 @@ target_include_directories(plist PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/PrivateHea
 install(TARGETS plist DESTINATION usr/lib)
```
Ktwu: Various fixes and additions to PlistBuddy: Onha should look as this.

Libraries/pbxbuild/Sources/Phase/File.cpp
grp: `ext::optional<T>` has an implicit constructor from `T`, so when the `std::string` was passed to the function taking `ext::optional<std::string>`, it was implicitly converted. (Note this commit already landed as #205.)

Libraries/ext/Headers/ext/optional
```diff
@@ -168,6 +168,14 @@ namespace std { namespace experimental { inline namespace fundamentals_v1 {
...
 #endif
 
+#if defined(__clang__) || defined(__GNUC__)
```
grp: Ya, I'd use that if it was standard. Otherwise it only helps for Clang which has supported these since before it had C++ support... so can't be bothered.

Libraries/graphics/Sources/Format/PNG.cpp
grp: All the code I've seen uses casts for it, so I assume down? I'm not quite sure I understand how this works but it's what the zlib tutorial did. ;P

Libraries/libbom/Sources/bom.c
grp: I had to remove the `first` array because it ended up with nested zero-length arrays, and 
MSVC complained about that. Since it was the last item in the `struct bom_variable`, adding the sizeof is equivalent to the old array location.

grp:
> Explicitly reference contents of the current namespace: what does this commit solve?

MSVC requires it. (I don't know if C++ also does and Clang/GCC are wrong, or if MSVC is wrong, but either way it fixes it in practice.)

Libraries/plist/Tools/PlistBuddy.cpp
grp: In `PlistBuddy` key specifiers can be `:CFBundleVersion` or just `CFBundleVersion`; i.e. the initial colon is optional. If it's there, then the contents up to the first `:` is empty, so don't pass that empty string along as a key name.

grp:
OK, remaining things to do here:

 - Let @onhachoe review the PlistBuddy commit.
 - Rebase & fix @ktwu's comments.

Did I miss anything?

Ktwu:
@grp Your list got everything.  I'll ask Onha to review the commit by EOD.

Libraries/plist/Tools/PlistBuddy.cpp
```diff
@@ -241,183 +216,235 @@ ParseCommandValueString(std::vector<std::string>::const_iterator begin, std::vec
 }
 
 static plist::Object *
-GetObjectAtKeyPath(plist::Object *object, std::queue<std::string> *remainingKeys, bool leaveLastKey = true)
+GetObjectAtKeyPath(plist::Object *object, std::queue<std::string> *keyPath, bool leaveLastKey = true)
```
onhachoe: This feels to me too much of code just to make recursion call to a iterative one...

grp: Hm it's now 35 lines but before it was 36 lines, seems about the same?

grp:
OK, fixed @Ktwu and @onhachoe comments and rebased. Only exception was @Ktwu 's comment about `_readObject()`, turns out there was already a `readObject()` so that's why it had an underscore.

grp:
@Ktwu @onhachoe, anything else needed?

PR #203, opened 2016-11-21
Add initial Windows support.

grp:
*This PR is built on #202, so it'd make sense for that to go in first.*

Adds basic Windows support. All tests pass, and many tools are usable; that said, I doubt you could run a full build without more work. Contents include:

 - Windows build configuration & CI. Builds and passes tests on Travis (Linux + MinGW + Wine) and AppVeyor (Windows + MSVC).
 - Requires zlib to be manually supplied. The CI builds download a precompiled build from zlib.net.
 - Replaces libpng and libxml2 with GDI+ and XmlLite on Windows. The API is very similar.
 - Platform libraries use Win32 instead of POSIX. For the most part these are interchangeable, but file permissions are different on Windows and not yet handled.

The major complicating factor is path names: Windows paths are vastly more complicated than Unix ones, and many previous assumptions don't hold true. The last few commits are a minimal refactor for basic support, but real support will involve replacing all string paths the path abstraction.


Libraries/xcworkspace/Tools/dump_xcworkspace.cpp
```diff
@@ -235,7 +235,7 @@ main(int argc, char **argv)
         std::sort(schemes.begin(), schemes.end(),
                 [](xcscheme::XC::Scheme::shared_ptr const &a, xcscheme::XC::Scheme::shared_ptr const &b) -> bool
                 {
-                    return ::strcasecmp(a->name().c_str(), b->name().c_str()) < 0;
+                    return ::strcmp(a->name().c_str(), b->name().c_str()) < 0;
```
grp: Yeah, this is just a debugging tool so doesn't really matter, and avoids the extra lines to be insensitive.

Libraries/libbom/Headers/bom/bom_format.h
```diff
@@ -30,6 +30,19 @@
 extern "C" {
 #endif
 
+#if _MSC_VER
```
grp: libbom is used standalone outside xcbuild so purposely has no dependencies. cc @marcsalem

Libraries/libbom/Headers/bom/bom_format.h
```diff
@@ -30,6 +30,19 @@
...
+#define __BOM_PACKED_STRUCT_BEGIN __pragma(pack(push, 1))
+#define __BOM_PACKED_STRUCT_END __pragma(pack(pop))
+#else
+#define __BOM_PACKED_STRUCT_BEGIN
+#define __BOM_PACKED_STRUCT_END __attribute__((packed))
+#endif
+
+#if _MSC_VER
+__pragma(warning(push))
+__pragma(warning(disable: 4200))
```
grp: Hm, I thought that was C99 but not C++, although that may not be true with C++11? Will need to look.

Either way, if the warning disable is needed, I'd prefer to do it inline since libbom is used externally. (Although not on Windows AFAIK.)

CMakeLists.txt
```diff
@@ -10,35 +10,69 @@
...
   set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib")
 endif ()
 
 # Output into root build directory.
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+# Language version.
```
grp: Ah, nice. Can you set that globally too?

Libraries/pbxbuild/Sources/Build/DependencyResolver.cpp
```diff
@@ -306,14 +306,13 @@ resolveSchemeDependencies(Build::Context const &context) const
...
-        DependenciesContext dependenciesContext = {
-            .buildEnvironment = &_buildEnvironment,
-            .buildContext     = &context,
-            .graph = &graph,
-            .buildAction = buildAction,
-            .positional = &positional,
-            .productNameToTarget = &productNameToTarget,
-        };
+        DependenciesContext dependenciesContext;
+        dependenciesContext.buildEnvironment = &_buildEnvironment;
```
grp: This struct is complicated enough I figured it was worth having the names inline rather than guessing at the order.

Libraries/pbxsetting/Sources/DefaultSettings.cpp
```diff
@@ -14,7 +14,9 @@
 
 #include <cstdlib>
 
+#if defined(__APPLE__)
```
grp: It's just for `confstr()` which is Darwin-only. Everything should go through the `process::Context`, but that isn't exposed there so the `confstr()` is accessed directly.

Libraries/plist/Sources/UnixTime.cpp
```diff
@@ -16,8 +16,13 @@ using plist::UnixTime;
 void UnixTime::
 Decode(uint64_t in, struct tm &out)
 {
+#if _WIN32
+    time_t const t = in;
```
grp: Needed to get it to compile, not sure why. I don't see `gmtime_s` in `man gmtime` on macOS, not sure about Linux. I think it's from a too-new version of C or POSIX or something? Maybe there's a C++ stdlib API for this I could use instead?

Libraries/plist/Sources/Format/BaseXMLParser.cpp
```diff
@@ -9,21 +9,284 @@
...
+#include <cstdarg>
+
+#if _WIN32
+#include <windows.h>
+#endif
 
 using plist::Format::BaseXMLParser;
 
+#if _WIN32
+using WideString = std::basic_string<std::remove_const<std::remove_pointer<LPCWSTR>::type>::type>;
```
grp: `std::wstring` exists on Unix too with different semantics; I'd prefer to avoid it completely than accidentally use it outside Windows.

I like `std::remove_pointer` and `std::remove_const` since they make it clear that the goal is a string of that type rather than assuming that `WCHAR` is the base, although I know it always is... yeah I guess there's no real point to it.

Libraries/libbom/Headers/bom/bom_format.h
sas: I'm not sure how that is a problem. I'll probably ask @marcsalem or @Ktwu.

CMakeLists.txt
sas: I'm not 100% sure. It might be a per-project thing. I don't think I actually know how CMake variables work exactly :p

Libraries/pbxbuild/Sources/Build/DependencyResolver.cpp
sas: Makes sense.

Libraries/plist/Sources/UnixTime.cpp
sas: Looks like `gmtime_s` is standard C11. Do we use C99 still?

Libraries/plist/Sources/Format/BaseXMLParser.cpp
sas: Yeah I suppose I don't disagree with the `remove_pointer` and `remove_const` stuff. If you want to keep this this way I'm fine with it.

Libraries/plist/Sources/UnixTime.cpp
grp: It appears that `gmtime_s` is part of an [optional][0] "secure programming" library in C11, detectable by the `__STDC_LIB_EXT1__` macro. Microsoft appears to have adopted them quite strongly (see *note*), but they seem to be [widely][1] [disliked][2] elsewhere so unlikely to appear on macOS or Linux.

[0]: http://stackoverflow.com/a/16700951
[1]: https://en.wikipedia.org/wiki/C11_(C_standard_revision)#Criticism
[2]: http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1967.htm

*Note*: By default, MSVC issues a warning for the "non-secure" variants unless `_CRT_SECURE_NO_WARNINGS` is defined; that's added through `CMakeLists.txt` in the build configuration patch.

Also, C++11 is based on C99. Only C++17 will change that to C11, so even if it was required in C11 it wouldn't necessarily be usable here.

grp: (All of this code could likely use `std::chrono` instead of the C time APIs, which would avoid the issue as well as being much more type safe.)

grp:
Ping!

Libraries/graphics/CMakeLists.txt
```diff
@@ -19,7 +19,9 @@ find_package(ZLIB REQUIRED)
 target_link_libraries(graphics PUBLIC ext "${ZLIB_LIBRARIES}")
 target_include_directories(graphics PRIVATE "${ZLIB_INCLUDE_DIR}")
 
-if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "Windows")
+  target_link_libraries(graphics PRIVATE wsock32 ws2_32)
```
grp: Not sure. Do you know what the difference is?

Libraries/libcar/Sources/AttributeList.cpp
```diff
@@ -48,7 +48,7 @@ dump() const
         uint16_t value = entry.second;
 
         if (identifier < sizeof(car_attribute_identifier_names) / sizeof(*car_attribute_identifier_names)) {
-            printf("[%02d] %-24s = %-6d | %-4x\n", identifier, car_attribute_identifier_names[identifier] ?: "(unknown)", value, value);
+            printf("[%02d] %-24s = %-6d | %-4x\n", identifier, car_attribute_identifier_names[identifier] ? car_attribute_identifier_names[identifier] : "(unknown)", value, value);
```
grp: It's unfortunate it's not standard!

CMakeLists.txt
grp: Looks like `CMAKE_CXX_STANDARD` allows setting a default, but is only available in CMake 3.1 and later. The minimum right now is CMake 3.0, but I think I can look into changing that later after this goes in.

Libraries/graphics/Sources/Format/PNG.cpp
```diff
@@ -10,19 +10,179 @@
 #include <graphics/Format/PNG.h>
 
 #include <iterator>
+#include <memory>
 #include <ext/optional>
 
 #if _WIN32
+#define CINTERFACE
```
grp: I sort of prefer having all of the `#define`s set in code, so the CMake file is less complicated. What's the harm in setting it in the few files that need it?

sas:
Closing this as I opened PR##251, a version that's rebased and with some fixes.

PR #205, opened 2016-12-04
Make fileNameDisambiguator optional in File::ResolveBuildFiles

mstorsjo:
Previously it was by default always set, but to an empty string.

This avoids having an empty outputBaseName later in ClangResolver,
which caused the output object files be named ".o" for all input
source files.


grp:
There's a version of this in #202 but good to get in even before that.

grp:
Thanks!

Issue #204, opened 2016-12-04
Support "-version Path" in xcodebuild


matthewbauer:
macOS xcodebuild seems to take options to "-version":

```
xcodebuild -version Path
```

This should output the SDK path.

This is seen in
https://github.com/qt/qtbase/blob/dev/configure#L271

## Update
This is actually mentioned in xcodebuild(1) man page:
```
xcodebuild -version [-sdk [sdkfullpath | sdkname]] [infoitem]
```

```
-version
Display version information for this install of Xcode. Does not
initiate a build. When used in conjunction with -sdk, the version
of the specified SDK is displayed, or all SDKs if -sdk is given no
argument.  Additionally, a single line of the reported version
information may be returned if infoitem is specified.
```

I cannot find any information on what "infoitem" but acceptable values seem to include:

* ProductVersion
* ProductBuildVersion
* Path

grp:
The usage you listed shows `-version -sdk <path>` which is implemented. Do you have an example command that fails?

grp:
However it looks like the `infoitem` parameters to `-version` aren't implemented. That would support printing out specific info about an SDK, I guess?

PR #206, opened 2016-12-05
Add support for printing specific information with version option.

grp:
Implements missing feature from #204.

### DIFF
--- a/Libraries/xcdriver/Sources/VersionAction.cpp
+++ b/Libraries/xcdriver/Sources/VersionAction.cpp
@@ -16,6 +16,8 @@
 #include <xcsdk/SDK/PlatformVersion.h>
 #include <xcsdk/SDK/Product.h>
 #include <xcsdk/SDK/Target.h>
+#include <plist/Dictionary.h>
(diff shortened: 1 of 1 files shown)


Libraries/xcdriver/Sources/VersionAction.cpp
```diff
@@ -60,36 +62,56 @@ Run(process::Context const *processContext, Filesystem const *filesystem, Option
...
+
+            fprintf(stdout, "\n");
+        } else {
+            for (std::string const &action : options.actions()) {
+                if (plist::String const *value = values->value<plist::String>(action)) {
+                    fprintf(stdout, "%s\n", value->value().c_str());
+                }
             }
         }
-        printf("\n");
```
Ktwu: Not sure why'd you move this into the empty actions block but w/e.

grp: By default it prints out the SDK then a blank line; with a specific item it doesn't print the blank line. I think that's to support piping the specific item output.

PR #207, opened 2016-12-05
added partial support for -p flags in lsbom

brtsai:


Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,68 @@ extern "C" {
...
+/**
+ *  Todo:
+ *      Group Name
+ *      PermissionsText
+ *      FileSizeFormatted
+ *      ModificationTimeFormatted
+ *      Username
+ *      UserGroupName
+*/
+std::string
```
grp: nit: `static std::string`

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,68 @@ extern "C" {
...
+printItemToString(Options::PrintItem item, std::string path, struct bom_path_info_2 *info){
```
grp: nit: `PrintItemToString` (or `StringFromPrintItem`)
nit: `{` on a new line

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,68 @@ extern "C" {
...
+
```
grp: nit: remove empty line

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,68 @@ extern "C" {
...
+            return std::to_string(ntohl(info->checksum));
+        case Options::PrintItem::FileName:
+            return path;
+        case Options::PrintItem::FileNameQuoted:
+            return "\"" + path + "\"";
+        case Options::PrintItem::GroupID:
+            return std::to_string(ntohl(info->group));
+        case Options::PrintItem::GroupName:
+
+        break;
```
grp: For these unfinished ones, I'd do it like this:

    case Options::PrintItem::SomeCase:
        return std::string(); // TODO: implement

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,68 @@ extern "C" {
...
+        case Options::PrintItem::UserGroupID:
+            return std::to_string(ntohl(info->user)) + "/" + std::to_string(ntohl(info->group));
+        break;
+        case Options::PrintItem::UserGroupName:
+
+        break;
+    }
+
+    return "";
```
grp: (Once the unimplemented ones have `return` above, can remove this.)

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -451,6 +513,18 @@ main(int argc, char **argv)
          */
         if (options->onlyPath()) {
             printf("%s\n", path.c_str());
+        } else if (options->printFormat()) {
+
```
grp: nit: remove empty line

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -451,6 +513,18 @@ main(int argc, char **argv)
          */
         if (options->onlyPath()) {
             printf("%s\n", path.c_str());
+        } else if (options->printFormat()) {
+            
+            bool first = true;
+            std::string toPrint;
+
+            for (Options::PrintItem item : *options->printFormat()) {
+                toPrint += first? "" : "\t";
```
grp: nit: space before `?`

brtsai:
screen shot 2016-12-05 at 4 19 09 pm

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,81 @@ extern "C" {
...
+{
+    std::vector<Options::PrintItem> toReturn;
+
+    toReturn.push_back(Options::PrintItem::FileName);
+    toReturn.push_back(Options::PrintItem::Permissions);
+    toReturn.push_back(Options::PrintItem::UserGroupID);
+    toReturn.push_back(Options::PrintItem::FileSize);
+    toReturn.push_back(Options::PrintItem::Checksum);
+
+    return toReturn;
```
grp: This could use an "initializer list" to create the vector:

    std::vector<Options::PrintItem> toReturn = {
        Options::PrintItem::FileName,
        Options::PrintItem::Permissions,
        Options::PrintItem::UserGroupID,
        Options::PrintItem::FileSize,
        Options::PrintItem::Checksum,
    };

    return toReturn;

Or even remove the temporary variable completely:

    return {
        Options::PrintItem::FileName,
        Options::PrintItem::Permissions,
        Options::PrintItem::UserGroupID,
        Options::PrintItem::FileSize,
        Options::PrintItem::Checksum,
    };

I like that style better for two reasons. Most importantly, I think it's easier to read: the structure of the code matches the structure of the data structure created. And not important in this case, but it's also slightly faster since it avoids calling `push_back` five times, each of which might need to expand the vector's allocated space.

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -449,25 +524,15 @@ main(int argc, char **argv)
...
-            printf("%s\t%s\t%s/%s", path.c_str(), mode.c_str(), UID.c_str(), GID.c_str());
-            
-            if (path_info_2_value->type == bom_path_type_file) {
-                std::string size = std::to_string(ntohl(path_info_2_value->size));
-                std::string checksum = std::to_string(ntohl(path_info_2_value->checksum));
-                printf("\t%s\t%s", size.c_str(), checksum.c_str());
-            }
-            
-            printf("\n");
+            toPrint = PrintItemsToString(GetDefaultPrintFormat(), path, path_info_2_value);
```
grp: Rather than the `else if` and `else`, here, you could do this:

    } else {
        std::vector<Options::PrintItem> printFormat = options->printFormat().value_or(GetDefaultPrintFormat());
        toPrint = PrintItemsToString(printFormat, path, path_info_2_value);
    }

Using the built-in [`value_or`](http://en.cppreference.com/w/cpp/experimental/optional/value_or) function to use a default if no `printFormat` is specified in the `options`.

Beyond that, you could even remove the `if` entirely:

    std::vector<Options::PrintItem> printFormat = options->printFormat().value_or(
        options->onlyPath() ? { Options::PrintItem::FileName  } : GetDefaultPrintFormat());
    toPrint = PrintItemsToString(printFormat, path, path_info_2_value);

Then, you could remove the `PrintItemsToString` function and just put the code right in here, since it's only called once.

Libraries/libbom/Tools/lsbom.cpp
```diff
@@ -294,6 +294,81 @@ extern "C" {
...
+            return std::string(); // TODO: implement
+        case Options::PrintItem::UserGroupID:
+            return std::to_string(ntohl(info->user)) + "/" + std::to_string(ntohl(info->group));
+        case Options::PrintItem::UserGroupName:
+            return std::string(); // TODO: implement
+    }
+}
+
+static std::vector<Options::PrintItem>
+GetDefaultPrintFormat ()
```
grp: nit: no space before `()`

Libraries/libbom/Tools/lsbom.cpp
brtsai: The initialiser list looks like a slightly faster, cleaner way to do this.
I'll make that change, thanks!

Libraries/libbom/Tools/lsbom.cpp
brtsai: I notice that in your value_or solution, if printFormat has been filled in via the -p flag, then onlyPath no longer takes precedence over the -p flag as lsbom.
Perhaps you meant:
`std::vector<Options::PrintItem> printFormat = options->onlyPath() ? std::vector<Options::PrintItem>({Options::PrintItem::FileName}) : options->printFormat().value_or(GetDefaultPrintFormat());`

And this too looks good!
I'll get this in there, and thanks again for the feedback!

grp: Good catch, sounds best to ignore the second half of my comment!

brtsai:
Tried pulling a fresh copy from the current main version, and noticed at lsbom doesn't seem to work anymore as I get "error: failed to load paths tree."

Tried tracing it to see what went wrong and here's what I got so far:

bom.c::bom_variable_iterate returned get_context.data_index == -1

	before bom.c::bom_variable_get returned -1

		before bom_tree.c::bom_tree_alloc_load returned null

			before bom_tree.c::bom_tree_alloc_load returned null
				
				before lsbom:: under comment Create Bom Tree for paths errored with “failed to load paths tree error”

grp:
@brtsai I think #225 will fix this

brtsai:
Thanks!
[225](https://github.com/facebook/xcbuild/pull/225) did it!
I've merged the latest ver into my repo, added my changes and pushed.
All of the tests seem to have passed and all of the changes posted in your review should be implemented.
Let me know if there's anything I've missed.

PR #211, opened 2016-12-06
add include for stdlib.h for abort()

onhachoe:
abort() requires #include <stdlib.h>


onhachoe:
Oooh.. the libbom things are completely of different matter, will strip that out, and make new change for that.

PR #213, opened 2016-12-06
Fix recursive createDirectory for relative paths

mstorsjo:
A path like "foo/bar/baz" will end up looking for a directory
named "" unless any of the earlier directories already exist.

### DIFF
--- a/Libraries/libutil/Sources/DefaultFilesystem.cpp
+++ b/Libraries/libutil/Sources/DefaultFilesystem.cpp
@@ -453,7 +453,7 @@ createDirectory(std::string const &path, bool recursive)
         std::stack<std::string> create;
 
         /* Build up list of directories to create. */
-        while (this->type(current) != Type::Directory) {
+        while (this->type(current) != Type::Directory && current != "") {
             create.push(current);
             current = FSUtil::GetDirectoryName(current);
         }


grp:
Hmm, do you know what hit this? All paths that come in here should be absolute; in the `MemoryFilesystem` it'll even fail if they aren't.

mstorsjo:
I run into this if I build with `SYMROOT=build`.

Normal xcodebuild (in my env at least) defaults to writing its outputs into a local directory named `build` instead of into `~/Library/Developer/Xcode/DerivedData` as when building from within xcode. With xcbuild, I seem to have to set this manually, in order to find the built products in the same place as they are when building with the native tools.

mstorsjo:
If it is assumed that all paths are absolute, I guess I could do `SYMROOT=$(pwd)/build` instead, although the plain relative path also works with the native tools.

grp:
For the needing to set `SYMROOT`, are you building a project or a workspace? If it's a project, are you specifying a scheme? (I haven't been quite sure when it should use `build/` versus DerivedData, but I think it's related do that.)

For the absolute paths, it should be xcbuild that resolves them to absolute paths before using them. Could you possibly get a backtrace of where this function is called with the relative path?

mstorsjo:
I'm building a project, and I'm not specifying a scheme, only a configuration.

Actually it seemed that `SYMROOT` didn't trigger this, setting `OBJROOT` does. Here's the relevant backtrace:

    #2  0x000000010072f1ff in libutil::DefaultFilesystem::createDirectory ()
    #3  0x000000010006d346 in xcexecution::SimpleExecutor::writeAuxiliaryFiles ()
    #4  0x000000010006bc1c in xcexecution::SimpleExecutor::buildTarget ()
    #5  0x000000010006b6c4 in xcexecution::SimpleExecutor::build ()
    #6  0x000000010001fc4a in xcdriver::BuildAction::Run ()
    #7  0x00000001000190ca in xcdriver::Driver::Run ()
    #8  0x0000000100005e94 in main ()

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -453,7 +453,7 @@ createDirectory(std::string const &path, bool recursive)
         std::stack<std::string> create;
 
         /* Build up list of directories to create. */
-        while (this->type(current) != Type::Directory) {
+        while (this->type(current) != Type::Directory && current != "") {
```
grp: nit: could be `!current.empty()`

grp:
OK, looked though the auxiliary files, and none of them resolve possible relative paths coming from build settings. Since that seems relatively difficult to fix before that code can be unit tested, this seems like a reasonable stopgap.

PR #217, opened 2016-12-09
Update Travis to a more recent macOS image.

grp:


grp:
Ping, since the holidays are over.

grp:
It's good for which C++ features are available in the compiler for building xcbuild, etc.

PR #218, opened 2016-12-09
Pass correct environment to tools.

grp:
- Include all necessary paths in tool search paths.
 - Set `PATH` to match the paths searched for tools.
 - Set `DEVELOPER_DIR` for all invocations to stay self-contained.
 - Pass along process environment to tools rather than clearing.

My version of #169. @onhachoe, would be great if you could test with the case you were seeing `PATH` being important.

### DIFF
--- a/Libraries/xcexecution/Sources/NinjaExecutor.cpp
+++ b/Libraries/xcexecution/Sources/NinjaExecutor.cpp
@@ -513,7 +513,7 @@ buildAction(
      * rules at the Ninja level. Instead, add a single rule that just passes through from
      * the build command that calls it.
      */
-    writer.rule(NinjaRuleName(), ninja::Value::Expression("cd $dir && env -i $env $exec && $depexec"));
+    writer.rule(NinjaRuleName(), ninja::Value::Expression("cd $dir && env $env $exec && $depexec"));
 
     /*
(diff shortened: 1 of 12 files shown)


Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -806,9 +806,9 @@ buildInvocation(
     }
 
     /*
-     * Build the invocation environment. To set the environment, we use standard shell syntax.
-     * Use `env` to avoid Bash-specific limitations on environment variables. Specifically, some
-     * versions of Bash don't allow setting "UID". Pass -i to clear out the environment.
+     * Build the invocation environment. To set the environment, we use standard shell tools:
+     * `env` to avoid Bash-specific limitations on environment variables (some versions of Bash
+     * don't allow setting "UID"). Intentionally add to, not replace, the process environment.
```
onhachoe: Is there a reason for changing the behavior?

grp: The old behavior was wrong, AFAICT.

onhachoe:
Wait, but how about other tools than ibtool actool? shouldn't we add support for scripts as well?

grp:
Scripts pass all build settings as environment variables, including `PATH`.

Issue #221, opened 2017-01-19
error: dereferencing type-punned pointer will break strict-aliasing rules


matthewbauer:
We are getting the following warning (being treated as an error) while building xcbuild with GCC compile:

```
/tmp/nix-build-xcbuild-0.1.1.drv-0/xcbuild-0.1.1-src/Libraries/plist/Sources/Format/ABPReader.cpp: In member function 'plist::Real* ABPReader::readReal(size_t)':
/tmp/nix-build-xcbuild-0.1.1.drv-0/xcbuild-0.1.1-src/Libraries/plist/Sources/Format/ABPReader.cpp:111:75: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
         case 4:  return plist::Real::New(*reinterpret_cast<float *>(&value)).release();
                                                                           ^
/tmp/nix-build-xcbuild-0.1.1.drv-0/xcbuild-0.1.1-src/Libraries/plist/Sources/Format/ABPReader.cpp:112:76: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
         case 8:  return plist::Real::New(*reinterpret_cast<double *>(&value)).release();
                                                                            ^
cc1plus: all warnings being treated as errors
make[2]: *** [Libraries/plist/CMakeFiles/plist.dir/build.make:687: Libraries/plist/CMakeFiles/plist.dir/Sources/Format/ABPReader.cpp.o] Error 1
```

Adding the flag -Wno-error=strict-aliasing fixes it.

grp:
@strager, do you know the best way to fix this?

strager:
Use memcpy instead of reinterpret_cast. memcpy (used in this way) results implementation-defined behaviour, but that's probably okay for now.

strager:
(The reinterpret_cast solution is undefined behaviour, so memcpy is strictly better.)

grp:
Fixed with #222.

PR #222, opened 2017-01-21
Use memcpy to transmute floating point values.

grp:
This is implementation-defined rather than undefined behavior.


grp:
Ping @strager / @Ktwu.

grp:
@strager I think you need to add yourself to the Facebook org before that works

grp:
@ktwu see #221

PR #225, opened 2017-01-31
Fix bom variable offset.

onhachoe:
Some of the vars->first were changed to vars + sizeof(struct bom_variable) in #202
where it really had to be sizeof(struct bom_variables)


grp:
Oops! Thanks for catching.

PR #233, opened 2017-02-09
[process] Capture stdout/stderr in DefaultLauncher

modocache:
== What's in this change? ==

* Modify the Launcher API such that it can return not only an exit code, but the stdout and stderr output as well.
* Modify DefaultLauncher such that it captures and returns stdout and stderr.

== Why make this change? ==

I'm working on adding an implementation of `swift-stdlib-tool` to xcbuild. This tool is expected to, among other things, scan object files in order to determine whether they contain references to the Swift runtime. In order to implement this logic in xcbuild, I could write C++ code that scans the object files, or I could use existing tools that do so. It's my intention to shell out to these existing tools, but
to do so I'd need to:

1. Find the appropriate tool by running `xcrun -find` and capturing its output.
2. Running the tool found by step (1) and capturing its output. I'd then scan the output for text that indicates a Swift runtime dependency.

And so I need a utility in xcbuild that can capture output from launched processes, which is what this change adds.

---

I haven't written very much C++, so please let me know if there's something I can improve upon here. Thanks!

### DIFF
--- a/Libraries/process/Headers/process/LaunchResult.h
+++ b/Libraries/process/Headers/process/LaunchResult.h
@@ -1,0 +1,50 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#ifndef __process_LaunchResult_h
+#define __process_LaunchResult_h
+
+#include <process/Context.h>
+
+#include <sstream>
+#include <ext/optional>
+
+namespace libutil { class Filesystem; }
+
+namespace process {
+
+/*
+ * The result of a successfully launched process: its exit code and output.
+ */
+class LaunchResult {
+private:
+    int _exitCode;
+    std::string _standardOutput;
+    std::string _standardError;
+
+public:
+    LaunchResult(
+        int exitCode,
+        std::string const &standardOutput,
+        std::string const &standardError);
+    ~LaunchResult();
+
+    int exitCode() const
+    { return _exitCode; }
+
+    std::string const &standardOutput() const
+    { return _standardOutput; }
+
+    std::string const &standardError() const
+    { return _standardError; }
+};
+
+}
+
+#endif  // !__process_LaunchResult_h
--- a/Libraries/process/Tests/test_DefaultLauncher.cpp
+++ b/Libraries/process/Tests/test_DefaultLauncher.cpp
@@ -1,0 +1,41 @@
+/**
+ Copyright (c) 2015-present, Facebook, Inc.
+ All rights reserved.
+
+ This source code is licensed under the BSD-style license found in the
+ LICENSE file in the root directory of this source tree. An additional grant
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <gtest/gtest.h>
+#include <process/DefaultLauncher.h>
+#include <process/LaunchResult.h>
+#include <process/MemoryContext.h>
+#include <libutil/DefaultFilesystem.h>
+
+#include <unistd.h>
+
+using process::DefaultLauncher;
+using process::MemoryContext;
+using libutil::DefaultFilesystem;
+
+TEST(DefaultLauncher, CapturesStdout)
+{
+    DefaultLauncher launcher;
+    auto filesystem = DefaultFilesystem::GetDefaultUNSAFE();
+
+    std::unordered_map<std::string, std::string> environment;
+    auto echo = MemoryContext(
+        "/bin/echo",
+        "/",
+        { "foo" },
+        std::unordered_map<std::string, std::string>(),
+        ::getuid(),
+        ::getgid(),
+        "user",
+        "group");
+    auto result = launcher.launch(filesystem, &echo);
+    EXPECT_TRUE(result);
+    EXPECT_EQ(0, result->exitCode());
+    EXPECT_EQ("foo\n", (*result).standardOutput());
+}
--- a/Libraries/xcexecution/Sources/NinjaExecutor.cpp
+++ b/Libraries/xcexecution/Sources/NinjaExecutor.cpp
@@ -21,6 +21,7 @@
 #include <process/Context.h>
 #include <process/MemoryContext.h>
 #include <process/Launcher.h>
+#include <process/LaunchResult.h>
 #include <libutil/md5.h>
 
 #include <sstream>
@@ -462,8 +463,8 @@ build(
             processContext->userName(),
             processContext->groupName());
 
-        ext::optional<int> exitCode = processLauncher->launch(filesystem, &ninja);
-        if (!exitCode || *exitCode != 0) {
+        auto result = processLauncher->launch(filesystem, &ninja);
+        if (!result || result->exitCode() != 0) {
             return false;
         }
     }
--- a/Libraries/xcexecution/Sources/SimpleExecutor.cpp
+++ b/Libraries/xcexecution/Sources/SimpleExecutor.cpp
@@ -18,6 +18,7 @@
 #include <process/Context.h>
 #include <process/MemoryContext.h>
 #include <process/Launcher.h>
+#include <process/LaunchResult.h>
 
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -291,8 +292,8 @@ performInvocations(
                         processContext->groupID(),
                         processContext->userName(),
                         processContext->groupName());
-                    ext::optional<int> exitCode = processLauncher->launch(filesystem, &context);
-                    success = (exitCode && *exitCode == 0);
+                    auto result = processLauncher->launch(filesystem, &context);
+                    success = (result && result->exitCode() == 0);
 
                     xcformatter::Formatter::Print(_formatter->finishInvocation(invocation, *path, createProductStructure));
                 } else {
--- a/Libraries/xcsdk/Tools/xcrun.cpp
+++ b/Libraries/xcsdk/Tools/xcrun.cpp
@@ -19,6 +19,7 @@
 #include <process/DefaultContext.h>
 #include <process/MemoryContext.h>
 #include <process/Launcher.h>
+#include <process/LaunchResult.h>
 #include <process/DefaultLauncher.h>
 #include <pbxsetting/Type.h>
 
@@ -451,13 +452,13 @@ static int Run(Filesystem *filesystem, process::Context const *processContext, p
                 processContext->userName(),
                 processContext->groupName());
 
-            ext::optional<int> exitCode = processLauncher->launch(filesystem, &context);
-            if (!exitCode) {
+            auto result = processLauncher->launch(filesystem, &context);
+            if (!result) {
                 fprintf(stderr, "error: unable to execute tool '%s'\n", options.tool()->c_str());
                 return -1;
             }
 
-            return *exitCode;
+            return result->exitCode();
         }
     }
 }
--- a/Libraries/process/CMakeLists.txt
+++ b/Libraries/process/CMakeLists.txt
@@ -12,6 +12,7 @@ add_library(process SHARED
             Sources/DefaultContext.cpp
             Sources/MemoryContext.cpp
             Sources/Launcher.cpp
+            Sources/LaunchResult.cpp
             Sources/DefaultLauncher.cpp
             Sources/MemoryLauncher.cpp
             )
@@ -22,3 +23,7 @@ install(TARGETS process DESTINATION usr/lib)
 
 find_package(Threads REQUIRED)
 target_link_libraries(process PRIVATE ${CMAKE_THREAD_LIBS_INIT})
+
+if (BUILD_TESTING)
+  ADD_UNIT_GTEST(process DefaultLauncher Tests/test_DefaultLauncher.cpp)
+endif ()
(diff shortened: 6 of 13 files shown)


Libraries/process/Headers/process/LaunchResult.h
```diff
@@ -0,0 +1,50 @@
...
+    std::string _stdoutOutput;
+    std::string _stderrOutput;
+
+public:
+    LaunchResult(int exitCode, std::string stdoutOutput, std::string stderrOutput):
+      _exitCode(exitCode), _stdoutOutput(stdoutOutput), _stderrOutput(stderrOutput) {}
+
+    int exitCode() const {
+      return _exitCode;
+    }
```
grp: nit: format like this:

    int exitCode() const
    { return _exitCode; }

Libraries/process/Headers/process/LaunchResult.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+class LaunchResult {
+private:
+    int _exitCode;
+    std::string _stdoutOutput;
+    std::string _stderrOutput;
+
+public:
+    LaunchResult(int exitCode, std::string stdoutOutput, std::string stderrOutput):
+      _exitCode(exitCode), _stdoutOutput(stdoutOutput), _stderrOutput(stderrOutput) {}
```
grp: nit: one line for each variable
nit: put this in the .cpp file instead

Libraries/process/Headers/process/LaunchResult.h
```diff
@@ -0,0 +1,50 @@
...
+ * The result of a successfully launched process: its exit code and output.
+ */
+class LaunchResult {
+private:
+    int _exitCode;
+    std::string _stdoutOutput;
+    std::string _stderrOutput;
+
+public:
+    LaunchResult(int exitCode, std::string stdoutOutput, std::string stderrOutput):
```
grp: nit: pass these strings as `std::string const &` to avoid extra copy

Libraries/process/Headers/process/LaunchResult.h
```diff
@@ -0,0 +1,50 @@
...
+      return _exitCode;
+    }
+
+    std::string const &stdoutOutput() const {
+      return _stdoutOutput;
+    }
+
+    std::string const &stderrOutput() const {
+      return _stderrOutput;
+    }
```
grp: (same for these two)

Libraries/process/Headers/process/LaunchResult.h
```diff
@@ -0,0 +1,50 @@
...
+namespace process {
+
+/*
+ * The result of a successfully launched process: its exit code and output.
+ */
+class LaunchResult {
+private:
+    int _exitCode;
+    std::string _stdoutOutput;
+    std::string _stderrOutput;
```
grp: nit: maybe call it `standardOutput` and `standardError`?

Libraries/process/Tests/test_DefaultLauncher.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::vector<std::string> arguments;
+    arguments.push_back("foo");
+    std::unordered_map<std::string, std::string> environment;
+    auto echo = MemoryContext(
+        "/bin/echo",
+        "/",
+        arguments,
+        std::unordered_map<std::string, std::string>(),
+        ::getuid(),
+        ::getgid(),
```
grp: Rather than use these which can change and make the test change in different contexts, maybe just use `0`?

Libraries/process/Tests/test_DefaultLauncher.cpp
```diff
@@ -0,0 +1,43 @@
...
+using process::MemoryContext;
+using libutil::DefaultFilesystem;
+
+TEST(DefaultLauncher, CapturesStdout)
+{
+    DefaultLauncher launcher;
+    auto filesystem = DefaultFilesystem::GetDefaultUNSAFE();
+
+    std::vector<std::string> arguments;
+    arguments.push_back("foo");
```
grp: Could just use `{ "foo" }` below or here.

Libraries/process/Tests/test_DefaultLauncher.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <unistd.h>
+
+using process::DefaultLauncher;
+using process::MemoryContext;
+using libutil::DefaultFilesystem;
+
+TEST(DefaultLauncher, CapturesStdout)
+{
+    DefaultLauncher launcher;
+    auto filesystem = DefaultFilesystem::GetDefaultUNSAFE();
```
grp: This isn't right — this means the test depends on the entire system filesystem. Instead, use a `MemoryFilesystem` to make it self-contained.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -462,8 +463,8 @@ build(
             processContext->userName(),
             processContext->groupName());
 
-        ext::optional<int> exitCode = processLauncher->launch(filesystem, &ninja);
-        if (!exitCode || *exitCode != 0) {
+        auto result = processLauncher->launch(filesystem, &ninja);
+        if (!result || (*result).exitCode() != 0) {
```
grp: nit: `result->exitCode()`

Libraries/xcexecution/Sources/SimpleExecutor.cpp
```diff
@@ -291,8 +292,8 @@ performInvocations(
                         processContext->groupID(),
                         processContext->userName(),
                         processContext->groupName());
-                    ext::optional<int> exitCode = processLauncher->launch(filesystem, &context);
-                    success = (exitCode && *exitCode == 0);
+                    auto result = processLauncher->launch(filesystem, &context);
+                    success = (result && (*result).exitCode() == 0);
```
grp: (same)

Libraries/xcsdk/Tools/xcrun.cpp
```diff
@@ -451,13 +452,13 @@ static int Run(Filesystem *filesystem, process::Context const *processContext, p
...
-            ext::optional<int> exitCode = processLauncher->launch(filesystem, &context);
-            if (!exitCode) {
+            auto result = processLauncher->launch(filesystem, &context);
+            if (!result) {
                 fprintf(stderr, "error: unable to execute tool '%s'\n", options.tool()->c_str());
                 return -1;
             }
 
-            return *exitCode;
+            return (*result).exitCode();
```
grp: same

grp:
Note I updated the comment above since I misinterpreted your plan originally.

modocache:
Thanks for the review, @grp! ❤️ 

I'm glad you agree that the ability to capture `stdout` and `stderr` is useful because, as I explain above, I not only need to find the path to a tool, I also need to parse that tool's output. Specifically, `swift-stdlib-tool` reads an executable or library's object files in order to determine whether Swift runtime libraries need to be copied to the location that the executable expects them to exist at. In order to accomplish this, I was thinking of shelling out to a tool such as `nm`, then scanning its output for Swift symbols. One approach would be to check whether symbols such as `__swift_FORCE_LOAD_$_swiftFoundation` exist, and if they do, copy `libswiftFoundation.dylib` as a result.

Alternatives to shelling out to `nm` and capturing its output could be:

1. Implementing object file parsing in xcbuild itself, or
2. Using a library that's capable of reading object files

I'd prefer not to do either, and instead shell out to `nm`. (1) involves reinventing the wheel a bit, by rewriting object parsing code that surely exists already elsewhere. I'd also like to implement `swift-stdlib-tool` such that it can read Mach-O, ELF, or COFF object files, which ends up being a non-trivial amount of work. On the other hand, (2) involves adding an additional dependency, probably on LLVM, for its object parsing code. I discussed these options with @onhachoe and we came to the conclusion  that shelling out to another tool such as `nm` was probably the simplest approach. Let me know if that makes sense to you.

> you can just use `filesystem->findExecutable()` to search those for the executable you actually need to run

Awesome! This sounds like it would also work, so I'll go with this approach. Thanks!

> Finally, I think this is missing actually printing the output in the cases that is needed.

I couldn't find any automated tests for these cases, but I did run a few commands that I thought fell into this category manually, in order to double-check before sending this PR. I'll try the cases you mentioned and report back.

I'll start on addressing some of the points you raised and update the PR.

Libraries/process/Tests/test_DefaultLauncher.cpp
modocache: Yeah, I had a bit of trouble here. I figured that I shouldn't use a method suffixed with `UNSAFE`, but I couldn't figure out how to use `MemoryFilesystem`. It seems like I can instantiate a `MemoryFilesystem` with files and directories, but not executables that would print to stdout. Other tests like `xcexecution/Tests/test_SimpleExecutor.cpp` use `process::MemoryLauncher` handlers, but that doesn't fit my use case, since I'd like to test `process::DefaultLauncher` here.

Another interesting approach taken in `test_SimpleExecutor.cpp` is using `builtin::Registry::Create` to instantiate executables, but that might not be possible here, since the `process` library would need to take a circular dependency on `builtin`. Or maybe I could simply have the `process` _tests_ dependent on `builtin`...?

Anyway, let me know if I'm missing something. I assume there's a better way that doesn't involve using a real filesystem, but I'm not sure what that is.

Libraries/process/Tests/test_DefaultLauncher.cpp
modocache: `0` might work when using a `MemoryFilesystem`. Using a `DefaultFilesystem` I get `setuid: Operation not permitted` and `setgid: Operation not permitted`. I'll definitely try this if I can figure out a way to use `MemoryFilesystem`, though!

modocache:
> This captures all output at once, but for tools with big outputs or that output over time, it's much better to stream that as it comes in.

Yeah, this is a really good point, and something I struggled with here. Is there a good way to achieve this? Maybe by adding stderr and stdout handler arguments to the `launch` method? That way people who call the method can determine what to do with output as it is streamed out.

Alternatively, I could rewrite this such that output is _both_ captured _and_ output to stdout and stderr as normal.

What do you think?

Libraries/process/Tests/test_DefaultLauncher.cpp
grp: Hmm, I think the issue is that you're trying to test the real (`Default`) implementations. Those are inherently not isolated since they are the bridge to the real OS, so they'd need to be integration tests. I think for now it might be best just not to test it until some kind of integration test infrastructure is in place?

grp:
> Let me know if that makes sense to you.

Agree, shelling out to `nm` seems like the easiest approach to start with. As long as it's using the same directories from the build to search out the `nm` tool, that should be just fine, and it shouldn't be hard to use the right directories since they're passed as arguments to `builtin-swiftStdlibTool`.

> I couldn't find any automated tests for these cases, but I did run a few commands that I thought fell into this category manually, in order to double-check before sending this PR. I'll try the cases you mentioned and report back.

Unfortunately there aren't tests for the output of `SimpleExecutor` or `NinjaExecutor`, exactly *because* there isn't a system in place yet to mock a process input and output to test it. I think this leads to the right solution: there's a missing abstraction for process I/O, similar to `process::Launcher` and `process::Context` for launching and context respectively.

I'm not sure what the API would look like exactly, but it'd have two implementations. (For now, I'll call it `process::Stream`.) One, a `process::DefaultStream` which reads and writes from the real `stdin`, `stderr`, `stdout`, etc. The other would be a `process::MemoryStream` that holds the data in memory.

Once the `process::Launcher` is adapted to return a stream, your code can use that to read the process output. For tests, the `process::MemoryLauncher` would return a `process::MemoryStream` that would be configured with predefined test output. In the normal case, the `process::DefaultLauncher` would hook up a `process::DefaultStream` to the real process's standard streams.

The `process::Stream` would also be useful in a few other places. For one, it'd allow testing the `SimpleExecutor` and `NinjaExecutor` output by passing in a `process::MemoryStream` that the executor writes to, then verifying the resulting contents after the execution. It'd also be useful in the `builtin` module, again to test the output of those pseudo-processes. And it could replace the global (i.e. non-testable) `isatty()` check for color output in `BuildAction.cpp`.

For the actual API, there's two concerns I can think of in advance. One is that it might need to have separate types for the input and output/error streams, since one is read-only and the other is write-only. The other concern is how the `process::Launcher` and `builtin::Driver` would work with streaming. Right now, their `run()` and `launch()` methods wait for the process to exit, but that doesn't work if you need to read and write interactively as the process runs. Instead, they might need to return immediately and expose a separate API to wait for the process to finish. That becomes more complex for the in-process implementations (`process::MemoryLauncher` and `builtin::Driver`) because without coroutines it's not easy to suspend the pseudo-process's code to wait for input. A special-purpose coroutine API and/or multithreading might work OK, though.

modocache:
Thanks @grp, this all makes sense to me. I'm having trouble with one part, though:

> The other concern is how the `process::Launcher` and `builtin::Driver` would work with streaming. Right now, their `run()` and `launch()` methods wait for the process to exit, but that doesn't work if you need to read and write interactively as the process runs. Instead, they might need to return immediately and expose a separate API to wait for the process to finish. That becomes more complex for the in-process implementations (`process::MemoryLauncher` and `builtin::Driver`) because without coroutines it's not easy to suspend the pseudo-process's code to wait for input. A special-purpose coroutine API and/or multithreading might work OK, though.

I may not have written enough C++ API to understand what this would look like. Can you think of an example of a similar API I could mimic? Or maybe comment with some pseudocode to give me an idea?

modocache:
Thanks for all the help, @grp! In the end I've made an implementation of swift-stdlib-tool in Python, outside of the xcbuild repository, that is working pretty well for my purposes. I think adding output streaming abstractions to xcbuild, and then building swift-stdlib-tool on top of that, would be a cool project for anyone who's interested, though!

PR #236, opened 2017-02-24
Fix xcspec to make swift compilation work properly

onhachoe:


Specifications/Compiler/com.apple.xcode.tools.swift.compiler.xcspec
```diff
@@ -108,6 +111,12 @@
             CommandLineFlag = "-sdk";
         },
         {
+            Name = "SWIFT_DEPLOYMENT_TARGET";
+            Type = String;
+            DefaultValue = "$($(DEPLOYMENT_TARGET_SETTING_NAME))";
+            CommandLineArgs = ( "-target", "$(CURRENT_ARCH)-apple-$(SWIFT_PLATFORM_TARGET_PREFIX)$(value)" );
```
modocache: It's a shame if neither the target nor the vendor ("apple" in this case) come down via some sort of environment variable, but I guess that's the case here?

PR #240, opened 2017-03-24
Get stdout/stderr of launched task with pipe.

onhachoe:
We didn't setup pipe to receive output of execve, which caused problems
as xcbuild and launched task would try to write to stdout/stderr unsynchronized.

Fix this by piping in output from child task, and manually printing them from xcbuild itself.

### DIFF
--- a/Libraries/process/Sources/DefaultLauncher.cpp
+++ b/Libraries/process/Sources/DefaultLauncher.cpp
@@ -10,8 +10,12 @@
 #include <process/DefaultLauncher.h>
 #include <libutil/Filesystem.h>
 
+#include <fcntl.h>
 #include <sys/wait.h>
 #include <unistd.h>
+
+// In most cases, size of pipe will be greater than one page,
+#define PIPE_BUFFER_SIZE 4096
 
 using process::DefaultLauncher;
 using libutil::Filesystem;
@@ -71,6 +75,14 @@ launch(Filesystem *filesystem, Context const *context)
     uid_t uid = context->userID();
     gid_t gid = context->groupID();
 
+    /* Setup parent-child stdout/stderr pipe. */
+    int pfd[2];
+    bool pipe_setup_success = true;
+    if (pipe(pfd) == -1) {
+        ::perror("pipe");
+        pipe_setup_success = false;
+    }
+
     /*
      * Fork new process.
      */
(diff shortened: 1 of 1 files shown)


Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -80,6 +88,18 @@ launch(Filesystem *filesystem, Context const *context)
         return ext::nullopt;
     } else if (pid == 0) {
         /* Fork succeeded, new process. */
+        /* Setup pipe to parent, redirecting both stdout and stderr */
+        if (pipe_setup_success) {
+            dup2(pfd[1], fileno(stdout));
+            dup2(pfd[1], fileno(stderr));
```
sas: You should use `STDOUT_FILENO` and `STDERR_FILENO` instead of `fileno(stdout)`.

Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -101,6 +121,24 @@ launch(Filesystem *filesystem, Context const *context)
         return ext::nullopt;
     } else {
         /* Fork succeeded, existing process. */
+        if (pipe_setup_success) {
+            close(pfd[1]);
+            /* Read child's stdout/stderr through pipe, and output stdout */
+            while (true) {
+                char pin;
+                int readlen = read(pfd[0], &pin, 1);
```
sas: You'll have pretty bad performance with this. Consider the following code instead:
```
#define BUF_SIZE 1024
char pin[BUF_SIZE];
int readlen = read(pfd[0], &pin, sizeof(pin));
if (readlen > 0) {
    fwrite(pin, readlen, 1, stdout);
} else {
    // ...
```

onhachoe: Yeah, I guess I was being lazy and reading one char at a time.
Given that this will try to read many bytes of output from child processes, it makes sense to buffer them.

Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -13,6 +13,9 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+// In most cases, size of pipe will be greater than one page,
+#define kPIPE_BUFFER_SIZE 4096
```
sas: The lowercase `k` seems weird for a constant like this, but if that's the style used in the project it's fine :/

Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -80,6 +91,18 @@ launch(Filesystem *filesystem, Context const *context)
...
         /* Fork succeeded, new process. */
+        /* Setup pipe to parent, redirecting both stdout and stderr */
+        if (pipe_setup_success) {
+            dup2(pfd[1], STDOUT_FILENO);
+            dup2(pfd[1], STDERR_FILENO);
+            close(pfd[0]);
+            close(pfd[1]);
+        } else {
+            /* No parent-child pipe setup, just ignore outputs from child */
+            fclose(stdout);
```
sas: `fclose()`ing `stdout` and `stderr` is a no-op here because these will get re-initialized by the child process after the call to `execve()`; without an `execve()`, you'd still have the underlying file descriptors open, so you still have the issue.

The proper way to ignore the output from the child is this:

```
close(STDOUT_FILENO);
close(STDERR_FILENO);
open("/dev/null", O_RDONLY);
open("/dev/null", O_WRONLY);
```

You can add `STDIN_FILENO` to the snippets above if you want to cancel out the input fd as well.

Issue #237, opened 2017-03-16
Support building mixed targets


vhbit:
Steps to reproduce:
```
git clone https://github.com/Quick/Nimble
cd Nimble
xcbuild -project Nimble.xcodeproj -scheme "Nimble-iOS" -configuration "Debug" -sdk "iphonesimulator"
```

at some point it fails with:

```
Nimble/Sources/NimbleObjectiveC/NMBStringify.m:2:9: fatal error: 'Nimble/Nimble-Swift.h' file not found
```

My guess would be that in case of mixed code Swift module has to be built first.
`xcbuild` was built from  9754643ae0c237593292af22b7332110268edb62

onhachoe:
I'm aware of this exact issue, and working on a fix right now, PR for it should come pretty soon.

onhachoe:
The relevant PR is merged, closing this issue.

vhbit:
Tested on `master` - the issue is still present. It seems `xcbuild` misses some steps and also too aggressive in parallelising them. 

First of all, it usually fails a couple of times when starting compiling Swift sources, it seems it happens because not all headers are copied/processed? In Xcode there are 2 actions shown before Swift compilation: "Write auxiliary files" and "Create product structure".

If I run build again and again at some point it goes through and builds Swift module but still fails in Obj-C part with the same error as before `fatal error: 'Nimble/Nimble-Swift.h' file not found`. 

From the quick look the difference with Xcode build is that the latter after building Swift does the following for all architectures:
```
Ditto /Users/vhbit/Library/Developer/Xcode/DerivedData/Nimble-geamnsituxelwdbwhhulkgurwaku/Build/Products/Debug-iphonesimulator/Nimble.framework/Headers/Nimble-Swift.h /Users/vhbit/Library/Developer/Xcode/DerivedData/Nimble-geamnsituxelwdbwhhulkgurwaku/Build/Intermediates/Nimble.build/Debug-iphonesimulator/Nimble-iOS.build/Objects-normal/i386/Nimble-Swift.h
    cd /Users/vhbit/projects/external/Nimble
    export PATH="/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/bin:/Applications/Xcode.app/Contents/Developer/usr/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
[...]

PR #242, opened 2017-03-30
Fixes to make swift project work correctly and fixes to ninja execution path.

onhachoe:
This includes 4 commits

Compile swift before any other clang invocation
 - Swift compilation first before objc compilation.
 - Fixes issue address on #237 

Slight refactoring of PhaseInvocation resolving step
 - as title

Refactor builtin tool resolving in NinjaExecutor
 - Fixes to make symlinked xcbuild work properly

Refactor auxiliary file creation path for ninja
 - Auxiliary file handling improvements.

Add Swift Standard Library tool xcspec
 - swift-stdlib-tool xcspec to copy over swift libraries

### DIFF
--- a/Libraries/pbxbuild/Sources/Phase/SwiftResolver.cpp
+++ b/Libraries/pbxbuild/Sources/Phase/SwiftResolver.cpp
@@ -64,7 +64,7 @@ ShouldBundleSwiftRuntime(Tool::Context const *toolContext, pbxsetting::Environme
      * Need a place to put the runtime libraries. If the product is not a wrapper, then
      * there isn't a frameworks directory to include Swift in.
      */
-    if (productType->isWrapper()) {
+    if (!productType->isWrapper()) {
         return false;
     }
 
--- a/Libraries/pbxbuild/Sources/Target/Environment.cpp
+++ b/Libraries/pbxbuild/Sources/Target/Environment.cpp
@@ -461,19 +461,22 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
     std::vector<std::string> variants = ResolveVariants(environment);
     environment.insertFront(ArchitecturesVariantsLevel(architectures, variants), false);
 
-    /* At the target level and below, the SDKROOT changes to always be a SDK path. */
-    environment.insertFront(pbxsetting::Level({
-        pbxsetting::Setting::Create("SDKROOT", sdk->path()),
-    }), false);
-
     /* Determine toolchains. Must be after the SDK levels are added, so they can be a fallback. */
     std::vector<xcsdk::SDK::Toolchain::shared_ptr> toolchains;
+    std::vector<std::string> effectiveToolchainPaths;
     for (std::string const &toolchainName : pbxsetting::Type::ParseList(environment.resolve("TOOLCHAINS"))) {
         if (xcsdk::SDK::Toolchain::shared_ptr toolchain = buildEnvironment.sdkManager()->findToolchain(toolchainName)) {
             // TODO: Apply toolchain override build settings.
             toolchains.push_back(toolchain);
+            effectiveToolchainPaths.push_back(toolchain->path());
         }
     }
+
+    environment.insertFront(pbxsetting::Level({
+        /* At the target level and below, the SDKROOT changes to always be a SDK path. */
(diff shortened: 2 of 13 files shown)


Specifications/Tool/com.apple.build-tools.swift-stdlib-tool.xcspec
```diff
@@ -0,0 +1,115 @@
...
+            Type = Enumeration;
+            Values = (
+                none,
+                verbose,
+            );
+            DefaultValue = "verbose";
+            CommandLineArgs = {
+                none = ();
+                verbose = (
+                    "--verbose",
```
modocache: I'm not super familiar with xcspec syntax, so I apologize if this is already true here, but keep in mind `swift-stdlib-tool` has two verbosity settings: `--verbose` (verbose output) and `--verbose --verbose` (debugging output). Does this allow for `--verbose` to be passed twice?

Libraries/pbxbuild/Sources/Phase/SwiftResolver.cpp
```diff
@@ -64,7 +64,7 @@ ShouldBundleSwiftRuntime(Tool::Context const *toolContext, pbxsetting::Environme
      * Need a place to put the runtime libraries. If the product is not a wrapper, then
      * there isn't a frameworks directory to include Swift in.
      */
-    if (productType->isWrapper()) {
+    if (!productType->isWrapper()) {
```
onhachoe: It did not work at all in the first place... :(

Specifications/Tool/com.apple.build-tools.swift-stdlib-tool.xcspec
onhachoe: oh, I did not realize that was a valid option,
I will update this with support for that.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -97,11 +97,11 @@ NinjaDescription(std::string const &description)
 }
 
 static std::string
-NinjaHash(std::string const &input)
+NinjaHashInternal(const char *data, size_t size)
```
sas: `char const *data`, I think.

Libraries/xcexecution/Sources/NinjaExecutor.cpp
```diff
@@ -114,6 +114,12 @@ NinjaHash(std::string const &input)
     return ss.str();
 }
 
+static std::string
+NinjaHash(std::string const &input)
```
sas: What's the point of this? From what I can see, the only caller for these functions actually calls the `Internal` version. Also, keeping the signature the same and not having an extra wrapper seems pretty easy in this case.

Libraries/pbxbuild/Sources/Target/Environment.cpp
```diff
@@ -461,20 +461,23 @@ Create(Build::Environment const &buildEnvironment, Build::Context const &buildCo
...
             // TODO: Apply toolchain override build settings.
             toolchains.push_back(toolchain);
+            effectiveToolchainPaths.push_back(toolchain->path());
         }
     }
 
+    environment.insertFront(pbxsetting::Level({
+        /* At the target level and below, the SDKROOT changes to always be a SDK path. */
+        pbxsetting::Setting::Create("SDKROOT", sdk->path()),
+        pbxsetting::Setting::Create("EFFECTIVE_TOOLCHAINS_DIRS", pbxsetting::Type::FormatList(effectiveToolchainPaths)),
```
sas: Does this just insert an empty variable?

onhachoe: This is being used by xcspec Specifications/Tool/com.apple.build-tools.swift-stdlib-tool.xcspec

Libraries/xcexecution/Sources/NinjaExecutor.cpp
onhachoe: I think they are the same, but okay, I think that is preferable

Libraries/xcexecution/Sources/NinjaExecutor.cpp
onhachoe: I might have been lazy with this,
Will change callers' signatures to solve this.

PR #245, opened 2017-04-12
Process contents inside child directories first in Filesystem->readDirectory call

onhachoe:
This fixes case when Filesystem->removeDirectory() call is invoked with recursive option set to true.


onhachoe:
updated tests to reflect the changes

sas:
Still good.

PR #256, opened 2017-07-06
actool fixes

Ktwu:
I tried using xcbuild's actool to compile an asset catalog for use by iphonesimulator apps.  It didn't work, but going by the stack traces I can get farther with these fixes:

- Allocated, but unused, memory should be cleared out
- BOM tree entries should be sorted according to binary key value.

The next step seems to involve compressing images with some cubic spline interpolation algorithm (if that's what CSI stands for in my current stack traces, anyway).


Libraries/libbom/Sources/bom.c
```diff
@@ -248,14 +248,19 @@ bom_index_reserve(struct bom_context *context, size_t count)
     struct bom_header *header = (struct bom_header *)context->memory.data;
 
     /* Insert space for extra indexes at the end of the currently allocated space. */
-    uint32_t index_point = ntohl(header->index_offset) + ntohl(header->index_length);
-    size_t new_index_length = ntohl(header->index_length) + sizeof(struct bom_index) * count;
+    size_t old_index_length = ntohl(header->index_length);
+    uint32_t index_point = ntohl(header->index_offset) + old_index_length;
     ptrdiff_t index_delta = sizeof(struct bom_index) * count;
+    size_t new_index_length = old_index_length + index_delta;
```
onhachoe: just caching values to temp vars?
I don't think there is any logic changes here, right?

Libraries/libbom/Sources/bom_tree.c
```diff
@@ -228,15 +228,51 @@ bom_tree_add(struct bom_tree_context *tree_context, const void *key, size_t key_
...
     }
 
-    size_t entry_index = ntohs(paths->count);
+    size_t last_index = ntohs(paths->count);
+    size_t entry_index = 0;
+    size_t start_range = 0;
+    size_t end_range = last_index;
 
-    /* Set the indexes for the inserted entry. */
+    while (start_range < end_range) {
```
onhachoe: Would be nice to have a comment here what this block is actually doing.

Libraries/libbom/Sources/bom.c
Ktwu: No logic change.  I just changed the variable names so that they were more clear to me while working on this (also, I needed access to different variables for the memset() below).

Libraries/libbom/Sources/bom_tree.c
Ktwu: Can do.  This is, if I remember correctly, supposed to find the position in the BOM tree that the new entry sorts to (since it needs to be sorted for fast lookup).

PR #257, opened 2017-07-15
Enable missing-prototypes compiler warning

strager:
Enable Clang's and GCC's -Wmissing-prototypes diagnostic and
fix issues it catches. This unconcovered some dead code, so
delete the dead code.


sas:
I'm fine with this, but it looks like it breaks the builds on travis.

sas:
This needs only a small fix in cmake before landing. Are you interested in fixing it?

PR #263, opened 2017-10-03
Fix various bugs related to windows execution.

onhachoe:


Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -141,15 +147,22 @@ launch(Filesystem *filesystem, Context const *context)
...
+    /* Get the exit code of the process */
+    DWORD status;
+    bool processFailed = GetExitCodeProcess(process.hProcess, &status);
+
+    /* Close process handles */
+    CloseHandle(process.hThread);
+    CloseHandle(process.hProcess);
 
-    return static_cast<int>(process.dwProcessId);
+    if (processFailed) {
```
sas: This is incorrect.

The return value of `GetExitCodeProcess()` is an integer that indicates whether or not `GetExitCodeProcess()` executed properly. It doesn't indicate anything about the child.

I think you want something like this:

```
if (GetExitCodeProcess(process.hProcess, &status) == 0) {
  // throw an exception, or assert, etc, because GetExciteCodeProcess failed.
}

[...]

return static_cast<int>(status);
```

Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -38,6 +38,61 @@ StringToWideString(std::string const &str)
     MultiByteToWideChar(CP_UTF8, 0, str.data(), (int)str.size(), &wide[0], size);
     return wide;
 }
+
+static WideString
+EscapedToken(const WideString &token)
```
sas: I don't understand what this function is trying to do and what your changes do.

Libraries/process/Sources/DefaultLauncher.cpp
```diff
@@ -55,60 +107,12 @@ ext::optional<int> DefaultLauncher::
...
-                    abort();
-                }
-            }
+    WideString executablePath = StringToWideString(context->executablePath());
 
-            arguments += static_cast<wchar_t>('"');
-        }
+    WideString arguments = EscapedToken(executablePath);
+    for (std::string const &argument : context->commandLineArguments()) {
+        arguments += static_cast<wchar_t>(' ');
```
sas: I think you want to do
```
arguments.push_back(L' ');
```
instead of that `static_cast`. The same applies for all the wchar_t you're trying to create.

Also, *technically*, the character type for `WideString` is `WCHAR`, not `wchar_t`.

PR #266, opened 2017-10-18
Fix windows lib casing

onhachoe:
Add missed casing errors on #265


onhachoe:
mingw has all libs in lowercases, which causes problem on linux :(

Issue #268, opened 2017-12-06
win32, bom_context_memory_file is broken


dkimitsa:
1. on windows it will not work on zero length files, so there is workaround should be introduced in forcing minsize to 1   
>An attempt to map a file with a length of 0 (zero) fails with an error code of ERROR_FILE_INVALID. Applications should test for files with a length of 0 (zero) and reject those files.

2.
https://github.com/facebook/xcbuild/blob/34e535ea7dc38fd77f4927c4b9cb387643c7988e/Libraries/libbom/Sources/bom_memory.c#L174
if it fails check should be done against null, not INVALID_HANDLE_VALUE

dkimitsa:
same thing is here 
https://github.com/facebook/xcbuild/blob/34e535ea7dc38fd77f4927c4b9cb387643c7988e/Libraries/libbom/Sources/bom_memory.c#L92

sas:
Fixed with PR #271. Thanks for the report!

PR #271, opened 2017-12-12
Windows fixes

sas:
These commits fix the issues reported in issue #268.

### DIFF
--- a/Libraries/libbom/Sources/bom_memory.c
+++ b/Libraries/libbom/Sources/bom_memory.c
@@ -89,7 +89,7 @@ _bom_context_memory_mremap(struct bom_context_memory *memory, size_t size)
     memory->size = size;
 
     context->mapping = CreateFileMapping(context->handle, NULL, (context->writeable ? PAGE_READWRITE : PAGE_READONLY), 0, size, NULL);
-    assert(context->mapping != INVALID_HANDLE_VALUE);
+    assert(context->mapping != NULL);
 
     memory->data = (void *)MapViewOfFile(context->mapping, (context->writeable ? FILE_MAP_WRITE | FILE_MAP_READ : FILE_MAP_READ), 0, 0, 0);
     assert(memory->data != NULL);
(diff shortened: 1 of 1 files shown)


dkimitsa:
sorry guys, extra check for zero doesn't fix the case when bom_context_memory_file is creating file which of course is zero byte length 
for ex. from here:
Libraries/acdriver/Sources/CompileAction.cpp 
struct bom_context_memory memory = bom_context_memory_file(path.c_str(), true, 0);

currently worked out it with such snipped at my end:
+    // FIXME: this is a dirty workaround as CreateFileMapping on zero lengh
+    // file will cause 1006 error as specified in doc
+    if (!minimum_size)
+        minimum_size = 1;
     if ((size_t)size < minimum_size) {
         size = (DWORD)minimum_size;

sas:
Do you mind uploading a PR with your change?

dkimitsa:
I had fixed it as quick workaround to run actool, in general fix looks as bellow, but it could side effects on other parts of app
```patch
diff --git a/Libraries/libbom/Sources/bom_memory.c b/Libraries/libbom/Sources/bom_memory.c
index bd8ea154..f4a6afe3 100755
--- a/Libraries/libbom/Sources/bom_memory.c
+++ b/Libraries/libbom/Sources/bom_memory.c
@@ -89,7 +89,7 @@ _bom_context_memory_mremap(struct bom_context_memory *memory, size_t size)
     memory->size = size;
 
     context->mapping = CreateFileMapping(context->handle, NULL, (context->writeable ? PAGE_READWRITE : PAGE_READONLY), 0, size, NULL);
-    assert(context->mapping != INVALID_HANDLE_VALUE);
+    assert(context->mapping);
 
     memory->data = (void *)MapViewOfFile(context->mapping, (context->writeable ? FILE_MAP_WRITE | FILE_MAP_READ : FILE_MAP_READ), 0, 0, 0);
     assert(memory->data != NULL);
@@ -156,6 +156,10 @@ bom_context_memory_file(const char *fn, bool writeable, size_t minimum_size)
         return invalid;
     }
 
+    // FIXME: this is a dirty workaround as CreateFileMapping on zero lengh
+    // file will cause 1006 error as specified in doc
+    if (!minimum_size)
+        minimum_size = 1;
     if ((size_t)size < minimum_size) {
         size = (DWORD)minimum_size;
 
@@ -171,7 +175,7 @@ bom_context_memory_file(const char *fn, bool writeable, size_t minimum_size)
     }
 
     HANDLE mapping = CreateFileMapping(handle, NULL, (writeable ? PAGE_READWRITE : PAGE_READONLY), 0, size, NULL);
-    if (mapping == INVALID_HANDLE_VALUE) {
+    if (mapping == NULL) {
         CloseHandle(handle);
         return invalid;
     }
```

PR #273, opened 2018-01-12
libcar: silence a -Wsign-compare warning

compnerd:
When building libcar for Windows, the array size calculation would be of type
`unsigned long long` (aka `size_t`).  However, `identifier` is of type
`enum car_attribute_identifier` (aka `int`).  This results in a warning due to
the sign mismatch.  Perform an explicit cast to the size type for the
comparision to silence the warning.

Take the opportunity to move the constant expression of the array length into a
variable for ease of readability.

### DIFF
--- a/Libraries/libcar/Sources/AttributeList.cpp
+++ b/Libraries/libcar/Sources/AttributeList.cpp
@@ -44,10 +44,12 @@ void AttributeList::
 dump() const
 {
     for (auto const &entry : _values) {
+        constexpr auto num_identifiers =
+            sizeof(car_attribute_identifier_names) / sizeof(*car_attribute_identifier_names);
         enum car_attribute_identifier identifier = entry.first;
         uint16_t value = entry.second;
 
-        if (identifier < sizeof(car_attribute_identifier_names) / sizeof(*car_attribute_identifier_names)) {
+        if (static_cast<decltype(num_identifiers)>(identifier) < num_identifiers) {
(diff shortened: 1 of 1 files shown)


Libraries/libcar/Sources/AttributeList.cpp
```diff
@@ -44,10 +44,12 @@ void AttributeList::
 dump() const
 {
     for (auto const &entry : _values) {
+        constexpr const auto identifiers =
```
Ktwu: LGTM; maybe make this num_identifiers instead for clarity

Libraries/libcar/Sources/AttributeList.cpp
```diff
@@ -44,10 +44,12 @@ void AttributeList::
 dump() const
 {
     for (auto const &entry : _values) {
+        constexpr const auto num_identifiers =
```
sas: `constexpr const` is redundant. I'd go for `const auto num_identifiers` because it won't make a difference in the generated code and easier to read, but `constexpr auto num_identifiers` is ok too if you prefer.

compnerd: Sure, changed to `constexpr auto`

sas: Thank you sir.

Ktwu:
Ergh, sorry for Travis blocking this.  We're working on this.

Issue #274, opened 2018-01-15
print "usage" for unknown arguments


umlaeute:
after installing `xcbuild` (via [homebrew](https://github.com/thii/homebrew-xcbuild))I was kind of lost: whatever I tried would result in an `unknown argument` error:

~~~sh
$ xcbuild -h
error: unknown argument -h
$ xcbuild --help
error: unknown argument --help
$ man xcbuild
No manual entry for xbuild
$
~~~

only after searching the sources(sic!), i discovered that in order to get help i must specify `-help` (with a single dash).
Now I understand that single-dash arguments are somewhat standard with macOS. Unfortunately I'm coming from a GNU world (with single-dash short options and double-dash long options).

I think it would be very nice if:
- any unknown argument would (in addition to triggering the `unknown argument %s` message) be redirected to `-usage`.
- include `xcbuild -help` in the usage examples (probably at the very end).

sas:
Agreed. I don't have time to deal with this right now, but if somebody ends up submitting a PR I'll gladly accept it.

PR #276, opened 2018-01-16
undefine major/minor macros to prevent build failures with newer glibc

umlaeute:
while fixing #274, I noticed that xcbuild fails to build from source, at least on an up-to-date Debian/sid system, with the following error

~~~
../Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h:35:13: error: In the GNU C Library, "major" is defined
 by <sys/sysmacros.h>. For historical compatibility, it is
 currently defined by <sys/types.h> as well, but we plan to
 remove this soon. To use "major", include <sys/sysmacros.h>
 directly. If you did not intend to use a system-defined macro
 "major", you should undefine it after including <sys/types.h>. [-Werror]
     int major() const
             ^~~~~~~~~                                                                                                                                                                                                                                                                                                                                                               
In file included from ../Libraries/xcassets/Headers/xcassets/Asset/LaunchImage.h:19:0,
                 from ../Libraries/acdriver/Headers/acdriver/Compile/LaunchImage.h:14,
                 from ../Libraries/acdriver/Sources/Compile/Asset.cpp:23:
../Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h:41:13: error: In the GNU C Library, "minor" is defined
 by <sys/sysmacros.h>. For historical compatibility, it is
 currently defined by <sys/types.h> as well, but we plan to
 remove this soon. To use "minor", include <sys/sysmacros.h>
 directly. If you did not intend to use a system-defined macro
 "minor", you should undefine it after including <sys/types.h>. [-Werror]
     int minor() const
~~~

This PR does what the compiler-error suggests: undefine the "minor" and "major" macros (if they are defined) before using these names, fixing the build failure.

### DIFF
--- a/Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h
+++ b/Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h
@@ -13,6 +13,13 @@
 #include <ext/optional>
 #include <string>
 
+#ifdef major
+# undef major
+#endif
+#ifdef minor
+# undef minor
+#endif
+
 namespace xcassets {
 namespace Slot {
 
--- a/Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h
+++ b/Libraries/xcassets/Headers/xcassets/Slot/SystemVersion.h
@@ -13,6 +13,10 @@
 #include <ext/optional>
 #include <string>
 
+/*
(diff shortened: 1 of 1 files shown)


sas:
Travis jobs failed because of infra failures on the Travis side it seems. I reran them.

PR #281, opened 2018-02-27
Add --allow-non-standard-behavior and --allow-image-type webp to actool

lacbs:
Adding the capability to pack WEBP images into compiled assets catalog.
Since this is not standard behavior, I'm adding it under `--allows-image-types webp`.
This way (`--allows-image-types`) also makes it easier to add future new extension types.
Since this is not a feature supported by Apple's Xcode.actool, I'm also adding `--allow-extensions` as well, that makes it explicit it's incompatible behavior.

### DIFF
--- a/Libraries/acdriver/Headers/acdriver/Compile/Output.h
+++ b/Libraries/acdriver/Headers/acdriver/Compile/Output.h
@@ -10,6 +10,7 @@
 #ifndef __acdriver_Compile_Output_h
 #define __acdriver_Compile_Output_h
 
+#include <acdriver/NonStandard.h>
 #include <plist/Dictionary.h>
 #include <car/Writer.h>
 
@@ -48,6 +49,7 @@ private:
 private:
     ext::optional<std::string>         _appIcon;
     ext::optional<std::string>         _launchImage;
+    NonStandard::ImageTypeSet          _allowedNonStandardImageTypes;
 
 private:
     ext::optional<car::Writer>         _car;
@@ -63,7 +65,8 @@ public:
         std::string const &root,
         Format format,
         ext::optional<std::string> const &appIcon,
-        ext::optional<std::string> const &launchImage);
+        ext::optional<std::string> const &launchImage,
+        NonStandard::ImageTypeSet const &allowedNonStandardImageTypes = NonStandard::ImageTypeSet());
 
 public:
     /*
@@ -90,6 +93,13 @@ public:
      */
     ext::optional<std::string> const &launchImage()
     { return _launchImage; }
+
+    /*
+     * List of non-standard image formats that are allowed into an image set
+     * when non-standard behavior are enabled.
+     */
+    NonStandard::ImageTypeSet const &allowedNonStandardImageTypes() const
+    { return _allowedNonStandardImageTypes; }
 
 public:
     /*
--- a/Libraries/acdriver/Headers/acdriver/Options.h
+++ b/Libraries/acdriver/Headers/acdriver/Options.h
@@ -10,6 +10,8 @@
 #ifndef __acdriver_Options_h
 #define __acdriver_Options_h
 
+#include <acdriver/NonStandard.h>
+#include <acdriver/Result.h>
 #include <libutil/Options.h>
 
 #include <string>
@@ -72,6 +74,10 @@ private:
     ext::optional<std::string> _filterForDeviceModel;
     ext::optional<std::string> _filterForDeviceOsVersion;
     ext::optional<std::string> _assetPackOutputSpecifications;
+
+private:
+    // Options not compatible with Apple's actool
+    NonStandard::ActoolOptions  _nonStandardOptions;
 
 public:
     Options();
@@ -151,6 +157,15 @@ public:
     { return _filterForDeviceOsVersion; }
     ext::optional<std::string> const &assetPackOutputSpecifications() const
     { return _assetPackOutputSpecifications; }
+
+public:
+    NonStandard::ActoolOptions const &nonStandardOptions() const
+    { return _nonStandardOptions; }
+
+public:
+
+    bool
+    isValid(Result *result) const;
 
 private:
     friend class libutil::Options;
--- a/Libraries/acdriver/Sources/Compile/ImageSet.cpp
+++ b/Libraries/acdriver/Sources/Compile/ImageSet.cpp
@@ -151,11 +151,29 @@ CompileAsset(
 
         format = car::Rendition::Data::Format::JPEG;
     } else {
-        result->normal(
-            Result::Severity::Error,
-            "unknown file type",
-            filename);
-        return false;
+        ext::optional<NonStandard::ImageType> type = NonStandard::ImageTypeFromFileExtension(FSUtil::GetFileExtension(filename));
+        if (!type) {
+            result->normal(
+                Result::Severity::Error,
+                "unknown file type",
+                filename);
+            return false;
+        }
+        if (!compileOutput->allowedNonStandardImageTypes().count(*type)) {
+            result->normal(
+                Result::Severity::Error,
+                "forbidden file type",
+                filename);
+            return false;
+        }
+        if (!filesystem->read(&pixels, filename)) {
+            result->normal(
+                Result::Severity::Error,
+                "unable to read image file",
+                filename);
+            return false;
+        }
+        format = NonStandard::ImageTypeToDataFormat(*type);
     }
 
     bool createFacet = false;
--- a/Libraries/acdriver/Sources/Compile/Output.cpp
+++ b/Libraries/acdriver/Sources/Compile/Output.cpp
@@ -29,12 +29,14 @@ Output(
     std::string const &root,
     Format format,
     ext::optional<std::string> const &appIcon,
-    ext::optional<std::string> const &launchImage) :
-    _root          (root),
-    _format        (format),
-    _appIcon       (appIcon),
-    _launchImage   (launchImage),
-    _additionalInfo(plist::Dictionary::New())
+    ext::optional<std::string> const &launchImage,
+    NonStandard::ImageTypeSet const &allowedNonStandardImageTypes) :
+    _root                        (root),
+    _format                      (format),
+    _appIcon                     (appIcon),
+    _launchImage                 (launchImage),
+    _allowedNonStandardImageTypes (allowedNonStandardImageTypes),
+    _additionalInfo              (plist::Dictionary::New())
 {
 }
 
@@ -44,4 +46,3 @@ AssetReference(xcassets::Asset::Asset const *asset)
     // TODO: include [] for each key
     return asset->path();
 }
-
--- a/Libraries/acdriver/Sources/CompileAction.cpp
+++ b/Libraries/acdriver/Sources/CompileAction.cpp
@@ -245,6 +245,10 @@ run(Filesystem *filesystem, Options const &options, Output *output, Result *resu
     // TODO: support all options
     WarnUnsupportedOptions(options, result);
 
+    if (!options.isValid(result)) {
+       return;
+    }
+
     /*
      * Determine format to output compiled assets.
      */
@@ -261,7 +265,8 @@ run(Filesystem *filesystem, Options const &options, Output *output, Result *resu
         *options.compile(),
         *outputFormat,
         options.appIcon(),
-        options.launchImage());
+        options.launchImage(),
+        options.nonStandardOptions().allowImageTypes());
 
     /*
      * If necessary, create output archive to write into.
--- a/Libraries/acdriver/Sources/Options.cpp
+++ b/Libraries/acdriver/Sources/Options.cpp
@@ -87,7 +87,20 @@ parseArgument(std::vector<std::string> const &args, std::vector<std::string>::co
     } else if (!arg.empty() && arg[0] != '-') {
         return libutil::Options::AppendCurrent<std::string>(&_inputs, arg);
     } else {
+        ext::optional<std::pair<bool, std::string>> nonStandardResult = _nonStandardOptions.parseArgument(args, it);
+        if (nonStandardResult) {
+            return *nonStandardResult;
+        }
         return std::make_pair(false, "unknown argument " + arg);
     }
 }
 
+
+bool Options::
+isValid(Result *result) const
+{
+    if (!nonStandardOptions().isValid(result)) {
+        return false;
+    }
+    return true;
+}
--- a/Libraries/libcar/Headers/car/Rendition.h
+++ b/Libraries/libcar/Headers/car/Rendition.h
@@ -44,12 +44,21 @@ public:
              * Raw data.
              */
             Data,
+            /*
+             * Non-standard formats
+             */
+            NON_STANDARD_WEBP,
         };
 
         /*
          * The size of a pixel in the format.
          */
         static size_t FormatSize(Format format);
+
+        /*
+         * If the format is saved as raw data, that requires additional headers
+         */
+        static bool FormatSavedAsRawData(Format format);
 
     private:
         std::vector<uint8_t> _data;
--- a/Libraries/libcar/Sources/Rendition.cpp
+++ b/Libraries/libcar/Sources/Rendition.cpp
@@ -50,7 +50,24 @@ FormatSize(Rendition::Data::Format format)
             return 2;
         case Format::JPEG:
         case Format::Data:
+        case Format::NON_STANDARD_WEBP:
             return 1;
+    }
+
+    abort();
+}
+
+bool Rendition::Data::
+FormatSavedAsRawData(Rendition::Data::Format format)
+{
+    switch (format) {
+        case Format::PremultipliedBGRA8:
+        case Format::PremultipliedGA8:
+            return false;
+        case Format::JPEG:
+        case Format::Data:
+        case Format::NON_STANDARD_WEBP:
+            return true;
     }
 
     abort();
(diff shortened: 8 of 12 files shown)


Libraries/acdriver/Headers/acdriver/Compile/Output.h
```diff
@@ -41,13 +42,29 @@ class Output {
         Folder,
     };
 
+    enum class ExtraImageType {
```
Ktwu: For anything like WEBP that's meant as an extension, I strongly recommend moving that logic into its own files as much as possible.

At the very least, knowledge about WEBP ought to go into its own file.  The ExtraImage* classes also into their own file.

Libraries/acdriver/Headers/acdriver/Compile/Output.h
```diff
@@ -91,6 +109,12 @@ class Output {
     ext::optional<std::string> const &launchImage()
     { return _launchImage; }
 
+    /*
+     * If an image with a non-standard format should be compile.
```
Ktwu: This sounds like an incomplete thought -- it's not a bool (which I associate with "if"), it's a list.  Maybe change to something like "List of non-standard image formats that are allowed

Libraries/acdriver/Headers/acdriver/Compile/Output.h
```diff
@@ -138,6 +162,12 @@ class Output {
      * The identifier of an asset for use in results.
      */
     static std::string AssetReference(xcassets::Asset::Asset const *asset);
+
+public:
+    /*
+     * Translates a type extension of an acceptable extra type to ExtraImageType enum.
+     */
+    static ext::optional<ExtraImageType> ExtraImageTypeFromExtension(std::string const &extension);
```
Ktwu: This function should live in the ExtraImageType class.

Libraries/acdriver/Sources/Compile/ImageSet.cpp
```diff
@@ -59,6 +59,16 @@ GenerateIdentifier(void) {
...
 }
 
+static car::Rendition::Data::Format
+ExtraImageTypeToDataFormat(Output::ExtraImageType type) {
+    switch (type) {
+        case Output::ExtraImageType::WEBP:
+            return car::Rendition::Data::Format::WEBP;
+    }
+
+    abort();
```
Ktwu: Do you need abort?  Compilation should catch incomplete switch cases -- yours looks correctly exhausted.

Libraries/acdriver/Sources/Compile/Output.cpp
```diff
@@ -29,12 +29,14 @@ Output(
...
     Format format,
     ext::optional<std::string> const &appIcon,
-    ext::optional<std::string> const &launchImage) :
-    _root          (root),
-    _format        (format),
-    _appIcon       (appIcon),
-    _launchImage   (launchImage),
-    _additionalInfo(plist::Dictionary::New())
+    ext::optional<std::string> const &launchImage,
+    ExtraImageTypeSet allowedExtraImageTypes) :
```
Ktwu: ExtraImageTypeSet is a nontrivial object -- pass by reference?

Libraries/acdriver/Sources/Options.cpp
```diff
@@ -84,6 +84,10 @@ parseArgument(std::vector<std::string> const &args, std::vector<std::string>::co
         return libutil::Options::Next<std::string>(&_filterForDeviceOsVersion, args, it);
     } else if (arg == "--asset-pack-output-specifications") {
         return libutil::Options::Next<std::string>(&_assetPackOutputSpecifications, args, it);
+    } else if (arg == "--allow-extensions") {
```
Ktwu: I'd honestly move this into their own function or the top of the parsing logic -- something to visually distinguish them as extensions.  Comment as well?

Libraries/acdriver/Headers/acdriver/Compile/Output.h
lacbs: You are right, I change the plans here and forgot to update the comments.

Libraries/acdriver/Sources/Compile/ImageSet.cpp
lacbs: I need it for MinGW (or windows build breaks)...

Libraries/acdriver/Sources/Compile/Output.cpp
lacbs: Thanks :D

Libraries/acdriver/Headers/acdriver/Compile/Output.h
lacbs: I'll move extension logic into it's own file as much as possible :)

Libraries/acdriver/Headers/acdriver/Compile/Output.h
lacbs: ExtraImageType class was a enum class, it can not define functions... :S

Libraries/acdriver/Headers/acdriver/Options.h
```diff
@@ -73,6 +74,10 @@ class Options {
     ext::optional<std::string> _filterForDeviceOsVersion;
     ext::optional<std::string> _assetPackOutputSpecifications;
 
+private:
+    // Options not compatible with Apple's actool
+    Extensions::ActoolOptions        _extensions;
```
Ktwu: You've got several misalignments going on in this diff; match the file?  _extensions should be aligned with all the other member names.

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,73 @@
...
+std::pair<bool, std::string> Extensions::ActoolOptions::
+parseArgument(std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it)
+{
+    std::string const &arg = **it;
+
+    if (arg == "--allow-extensions") {
+        return libutil::Options::Current<bool>(&_allowExtensions, arg);
+    } else if (arg == "--allow-image-types") {
+        return InsertNextImageType(&_allowImageTypes, args, it);
+    } else {
```
Ktwu: Nit: it's not the extension's responsibility decide what happens if an argument can't be parsed.  Make this parseArgument return a nullable, check for it in the main options file, and return the appropriate thing there.

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,73 @@
...
+ of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <acdriver/Extensions.h>
+#include <libutil/Options.h>
+
+using acdriver::Extensions;
+
+static std::pair<bool, std::string>
+InsertNextImageType(Extensions::ImageTypeSet *imageTypeSet, std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it)
```
Ktwu: Just make imageTypeSet a reference, not a pointer.

lacbs:
@sas @strager, can you take a look at this?

lacbs:
Just checking, this shows "Ktwu requested changes", but I addressed it in the latest version.

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -239,12 +239,27 @@ WarnUnsupportedOptions(Options const &options, Result *result)
     }
 }
 
+static bool
+CheckInvalidOptions(Options const &options, Result *result)
```
Ktwu: Nit:  Name this to something that's more obviously boolsy like "areOptionsValid" -- otherwise it's not very clear to me what it means if it returns true (is the functiuon checking for the presence of invalid options but it didn't find any, or is the check supposed to fail *if* invalid options are present?).

Libraries/acdriver/Headers/acdriver/Extensions.h
```diff
@@ -0,0 +1,84 @@
...
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+namespace acdriver {
+
+/*
+ * Contains non-standard related logic
+ */
+class Extensions {
```
strager: I'm confused why there's a class (a namespace, really) called Extensions. Why group all extensions together? What do different (future) extensions have to do with each other?

strager: The term "extension" confused me when reading the code in ImageSet.cpp. "Extension" can refer to a *file* extension (like `.png`).

Libraries/acdriver/Sources/Compile/ImageSet.cpp
```diff
@@ -151,11 +151,29 @@ CompileAsset(
...
+            return false;
+        }
+        if (!filesystem->read(&pixels, filename)) {
+            result->normal(
+                Result::Severity::Error,
+                "unable to read image file",
+                filename);
+            return false;
+        }
+        format = Extensions::ImageTypeToDataFormat(*type);
```
strager: Why are extension formats any different from JPEG and PNG (lines 100-152)? Why not write the WebP-handling code like this to match how JPEG files are handled:

```cpp
     } else if (FSUtil::IsFileExtension(filename, "webp", true)) {
         if (!compileOutput->isWebpAllowed()) {
             result->normal(
                 Result::Severity::Error,
                 "forbidden file type",
                 filename);
             return false;
         }
         if (!filesystem->read(&pixels, filename)) {
             result->normal(
                 Result::Severity::Error,
                 "unable to read WebP file",
                 filename);
             return false;
         }
 
         format = car::Rendition::Data::Format::WEBP;
    } else {
 ``

File: Libraries/acdriver/Sources/Extensions.cpp [hunk 13]
```diff
@@ -0,0 +1,74 @@
...
+}
+
+ext::optional<std::pair<bool, std::string>> Extensions::ActoolOptions::
+parseArgument(std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it)
+{
+    std::string const &arg = **it;
+
+    if (arg == "--allow-extensions") {
+        return libutil::Options::Current<bool>(&_allowExtensions, arg);
+    } else if (arg == "--allow-image-types") {
```
strager: Why is the option called --allow-image-types if it accepts only one image type? Shouldn't this be called --allow-image-type (or --allow-webp)?

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,74 @@
...
+        return InsertNextImageType(_allowImageTypes, args, it);
+    } else {
+        return ext::nullopt;
+    }
+}
+
+bool Extensions::ActoolOptions::
+areOptionsValid(acdriver::Result *result) const
+{
+    if (!allowExtensions() && allowImageTypes().size() > 0) {
```
strager: Nit: Use !.empty() instead of .size()>0.

```cpp
    if (!allowExtensions() && !allowImageTypes().empty()) { 
```

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    abort();
+}
+
+ext::optional<Extensions::ImageType> Extensions::
+ImageTypeFromFileExtension(std::string const &extension)
+{
+    std::string lowerExtension;
+    std::transform(extension.begin(), extension.end(), std::back_inserter(lowerExtension), ::tolower);
```
strager: Could we use Boost's `iequals` function instead? https://stackoverflow.com/a/315463

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,74 @@
...
+    auto ret = libutil::Options::Next<std::string>(&value, args, it);
+    if (!ret.first) {
+        return ret;
+    }
+    ext::optional<Extensions::ImageType> type = Extensions::ImageTypeFromFileExtension(*value);
+    if (!type) {
+        return std::make_pair(false, "unknown image type " + *value);
+    }
+    imageTypeSet.insert(*type);
+    return std::make_pair(true, std::string());
```
strager: Nit: Use direct initialization?

```cpp
    return {true, {}};
```

Libraries/acdriver/Sources/Extensions.cpp
```diff
@@ -0,0 +1,74 @@
...
+InsertNextImageType(Extensions::ImageTypeSet &imageTypeSet, std::vector<std::string> const &args, std::vector<std::string>::const_iterator *it)
+{
+    ext::optional<std::string> value;
+    auto ret = libutil::Options::Next<std::string>(&value, args, it);
+    if (!ret.first) {
+        return ret;
+    }
+    ext::optional<Extensions::ImageType> type = Extensions::ImageTypeFromFileExtension(*value);
+    if (!type) {
+        return std::make_pair(false, "unknown image type " + *value);
```
strager: Why is `--allow-image-types png` not allowed?

I think the --allow-image-types interface is clunky. What about `--allow-webp-images`?

Libraries/acdriver/Headers/acdriver/Extensions.h
```diff
@@ -0,0 +1,84 @@
...
+    class ImageTypeHash
+    {
+    public:
+        std::size_t operator()(ImageType t) const
+        {
+            return static_cast<std::size_t>(t);
+        }
+    };
+
+    typedef std::unordered_set<ImageType, ImageTypeHash> ImageTypeSet;
```
strager: Why is this an std::unoreder_set? That seems pretty heavy-weight. Why not a plain C struct containing boolean members?

```cpp
struct ImageTypeSet {
    bool webp = false;
};
```

Libraries/acdriver/Sources/CompileAction.cpp
```diff
@@ -239,12 +239,25 @@ WarnUnsupportedOptions(Options const &options, Result *result)
     }
 }
 
+static bool
+AreOptionsValid(Options const &options, Result *result)
```
strager: Why isn't this a member function in the `Options` class?

Libraries/libcar/Headers/car/Rendition.h
```diff
@@ -44,13 +44,22 @@ class Rendition {
              * Raw data.
              */
             Data,
+            /*
+             * Extensions format
```
strager: I think this comment is unhelpful. What about `Raw data, WEBP format.` or deleting the comment entirely?

Libraries/libcar/Headers/car/Rendition.h
```diff
@@ -44,13 +44,22 @@ class Rendition {
...
+             */
+            EXTENSIONS_WEBP,
```
strager: Why is this called `EXTENSIONS_WEBP` and not just `WEBP`?

Libraries/libcar/Sources/Rendition.cpp
```diff
@@ -653,8 +674,7 @@ write() const
     output_bytes += sizeof(struct car_rendition_info_bytes_per_row);
 
     /* JPEG/RAW format adds one more header at the end */
```
strager: Update this comment?

Libraries/acdriver/Headers/acdriver/Extensions.h
lacbs: I just wanted to make it clear what's standard behavior (compatible with Apple's actool) and what's extra feature this provides.
Initially I mixed all together, but @Ktwu asked me to move this logic to its own file as much as possible. I think it's become much better like this.
I even used a class instead of namespace to make it clear when we are using that. (e.g. you can not do `using acdriver::Extensions::ImageTypeFromFileExtension` and then use `ImageTypeFromFileExtension` alone, you need to always call like `Extensions::ImageTypeFromFileExtension`)

Yeah, if you have another suggestion of name, that's why I tried to always use "fileExtension" when applicable.

Libraries/acdriver/Sources/Compile/ImageSet.cpp
lacbs: PNG is handled different from everything else, but the difference with jpeg is just that's one if standard, the other one is not...

Libraries/acdriver/Sources/Extensions.cpp
lacbs: You are right, this one I'll change. I don't want to use `--allow-webp` because `--allow-image-type` will make it easier if we want to add new types.

Libraries/acdriver/Sources/Extensions.cpp
lacbs: Will do :)

Libraries/acdriver/Sources/Extensions.cpp
lacbs: I don't think Boost is used in xcbuild, and I don't think this is enough for it to be added... Besides, this method way to compare is already used in xcbuild.

Libraries/acdriver/Headers/acdriver/Extensions.h
lacbs: I agree it's heavy... but I still prefer unordered_set with that enum key just to make it easier to add new file types...
That structure is not going to be used that much...
I can still change it, and have member functions to translate the enums if you like it...

Libraries/acdriver/Sources/CompileAction.cpp
lacbs: it could be :D

Libraries/libcar/Headers/car/Rendition.h
lacbs: Well, if I say "non-standard formats" would that be more helpful?

Libraries/libcar/Headers/car/Rendition.h
lacbs: To make it clear it's non-standard

Libraries/libcar/Sources/Rendition.cpp
lacbs: will do

Libraries/acdriver/Headers/acdriver/Extensions.h
lacbs: I can do `Extensions` -> `NonStandard` with everything in this diff. What do you think @strager and @Ktwu ?

lacbs:
I'll wait for feedback on renaming `Extensions` -> `NonStandard` before proceeding.

PR #282, opened 2018-03-09
Allow symlinked paths when passing an SDK to xcrun

Ktwu:
Several changes here:

- Filesystem::readSymbolicLink() now behaves like "readlink -f", so links will be
  thoroughly resolved.
- The SDK Manager will now try to resolve toolchains and SDKs passed its
  way using a FS object.  Most locations have been stubbed with a
  nullptr for the FS since those are internally-provided toolchain names
  / paths, but some locations that can take in user-provided input are
  different.
- I added some functionality to convert a path into a vector of
  components; I use this to iterate though a given path and invoke
  "readlink()" per step.

### DIFF
--- a/Libraries/libutil/Sources/DefaultFilesystem.cpp
+++ b/Libraries/libutil/Sources/DefaultFilesystem.cpp
@@ -9,6 +9,7 @@
 
 #include <libutil/DefaultFilesystem.h>
 #include <libutil/FSUtil.h>
+#include <libutil/Relative.h>
 
 #include <stack>
 #include <climits>
@@ -28,6 +29,8 @@
 #include <copyfile.h>
 #endif
 #endif
+
+#include <iostream>
 
 using libutil::DefaultFilesystem;
 using libutil::Filesystem;
@@ -678,6 +681,46 @@ removeFile(std::string const &path)
 
     return true;
 #endif
+}
+
+ext::optional<std::string> DefaultFilesystem::
+readSymbolicLinkCanonical(std::string const &path, bool *directory) const
+{
+    /* ::readlink() is not the equivalent of "readlink -f", which recursively
+       expands the path. Furthermore, it does not successfully expand a symlink
+       if only part of the path is symbolic -- so long as the path points to a
+       real location, the invocation will fail.
+
+       If a path like /a/b/c exists, where:
+         - /a -> /foo
+         - /foo/b -> /bar/
+         - /bar/c is the real file
+
+       Expanding the path out fully must operate on each individual component.
+    */
+    std::vector<std::string> components = FSUtil::NormalizePathComponents(path);
+    std::string resolvedPath;
+    std::string pathToResolve;
+    ext::optional<std::string> current;
+
+    for (auto it = components.begin(); it != components.end(); ++it) {
+        pathToResolve = libutil::Path::Relative(resolvedPath).child(*it).raw();
+        do {
+            current = this->readSymbolicLink(pathToResolve, directory);
+            if (current) {
+                resolvedPath = *current;
+            }
+        } while (current);
+    }
+
+    if (resolvedPath.empty()) {
+        return ext::nullopt;
+    }
+
+    if (directory) {
+        *directory = (this->type(this->resolvePath(resolvedPath)) == Type::Directory);
+    }
+    return std::string(resolvedPath.c_str());
 }
 
 ext::optional<std::string> DefaultFilesystem::
--- a/Libraries/libutil/Sources/Relative.cpp
+++ b/Libraries/libutil/Sources/Relative.cpp
@@ -30,8 +30,8 @@ BaseRelative(std::string const &raw) :
 {
 }
 
-static std::string
-NormalizePath(
+static std::vector<std::string>
+NormalizePathComponents(
     std::string const &path,
     std::string::const_iterator start,
     std::string::const_iterator end,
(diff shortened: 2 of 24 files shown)


strager:
Is it possible to bump the minimum C++ version to C++17? C++17 added a bunch of filesystem APIs which would solve some of the problems you're solving manually. For example, std::filesystem::canonical: http://en.cppreference.com/w/cpp/filesystem/canonical

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -730,23 +733,53 @@ readSymbolicLink(std::string const &path, bool *directory) const
     WideString target = WideString(&pathBuffer[offset], &pathBuffer[offset + length]);
     return WideStringToString(target);
 #else
-    for (size_t size = PATH_MAX; true; size *= 2) {
```
strager: I suggest put this code in its own function and having the iterative logic use that function.

In other words, keep `readSymbolicLink` as-is and introduce `resolveSymbolicLinks` which calls `readSymbolicLink`.

Right now, your logic isn't triggered on Windows (should it?), and I think that's an oversight. Also, this function is pretty complex (doubly nested loops with early returns) after your changes.

Libraries/libutil/Sources/Relative.cpp
```diff
@@ -106,7 +140,24 @@ normalized() const
         output += (Traits::IsSeparator(_raw[n]) ? Traits::Separator : _raw[n]);
     }
 
-    output += NormalizePath(_raw, _raw.begin() + start, _raw.end(), absolute, true, Traits::Separator, &Traits::IsSeparator);
+    std::vector<std::string> relativeOutput = NormalizePathComponents(
```
strager: Why does `normalized` not just call `normalizedComponents`? It looks like the functions are duplicates of each other aside from the loop on lines 153-160 and the conversion on lines 122-123.

Libraries/libutil/Sources/Relative.cpp
```diff
@@ -106,7 +140,24 @@ normalized() const
...
+    );
+
+    auto begin = relativeOutput.begin();
+    auto end = relativeOutput.end();
+    for (auto it = begin; it != end; ++it) {
+        output += *it;
+        if (std::next(it) != end) {
+          output += Traits::Separator;
+        }
+    }
```
strager: Put this loop into its own function? (Maybe that function already exists?)

Libraries/libutil/Tests/test_FSUtil.cpp
```diff
@@ -15,6 +15,15 @@ using libutil::FSUtil;
 // TODO: remove these tests once FSUtil is unused
 #if !_WIN32
 
+template <typename T>
+void EXPECT_VECTOR_EQ(std::vector<T> x, std::vector<T> y) {
```
strager: Nit: Accept a reference, not a copy:

```
void EXPECT_VECTOR_EQ(std::vector<T> const &x, std::vector<T> const &y) {
```

Libraries/libutil/Tests/test_FSUtil.cpp
```diff
@@ -15,6 +15,15 @@ using libutil::FSUtil;
...
+  ASSERT_EQ(x.size(), y.size()) << "Vectors x and y are of unequal length";
+
+  for (int i = 0; i < x.size(); ++i) {
+    EXPECT_EQ(x[i], y[i]) << "Vectors x and y differ at index " << i;
+  }
+}
```
strager: Nice!

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
```diff
@@ -61,14 +62,18 @@ class Manager {
     /*
      * Find an SDK by name. This does a fuzzy search, so the "name" could
      * be an SDK name or path, or even the name of a platform.
+     * The provided filesystem will attempt to resolve the name if the name is
+     * a symlinked path.
      */
-    Target::shared_ptr findTarget(std::string const &name) const;
+    Target::shared_ptr findTarget(libutil::Filesystem const *filesystem, std::string const &name) const;
```
strager: Why is the filesystem optional?

strager: Ah, you answered that in your commit message:

>  Most locations have been stubbed with a
> nullptr for the FS since those are internally-provided toolchain names
> / paths, but some locations that can take in user-provided input are
> different.

I don't understand this, but it sounds like you've thought about it already and I don't have a good reason for the filesystem to be required.

Libraries/libutil/Sources/Relative.cpp
```diff
@@ -40,56 +40,90 @@ NormalizePath(
...
-            }
-        } else if (collapse && (it == start || isSeparator(*std::prev(it))) && *it == '.') {
-            /* Handle relative paths. */
-            if (std::next(it) == end || isSeparator(*std::next(it))) {
-                /* Path component is current. */
-                if (!output.empty() && isSeparator(output.back())) {
+  std::vector<std::string> output;
+  std::string component;
+
+  for (auto it = start; it != end; ++it) {
```
strager: It looks like you screwed up the indentation. Undo these whitespace changes?

Libraries/libutil/Sources/DefaultFilesystem.cpp
strager: (As written, reviewing the code thoroughly will make my head hurt.)

Libraries/libutil/Sources/DefaultFilesystem.cpp
```diff
@@ -730,23 +733,53 @@ readSymbolicLink(std::string const &path, bool *directory) const
...
+            ++it;
         }
     }
+
+    if (directory) {
+        *directory = (this->type(this->resolvePath(resolvedPath)) == Type::Directory);
+    }
+    return std::string(resolvedPath.c_str());
 #endif
 }
```
strager: No tests? =O

Libraries/libutil/Sources/Relative.cpp
Ktwu: I just wanted to minimize changes to the code, but that doesn't make sense now that you point this duplication out.

Libraries/libutil/Sources/Relative.cpp
Ktwu: I don't think it does.  I'll check.

Libraries/xcsdk/Headers/xcsdk/SDK/Manager.h
Ktwu: Ultimately, I'm just using the filesystem to resolve symlinks.  If no filesystem, no symlink resolution, but for internal invocations that's OK since xcbuild won't call this internally with symlinks.  Only user-facing tools could possibly pass a symlink (from what I've read of the code, anyway).

Libraries/libutil/Sources/Relative.cpp
Ktwu: Sure.

Libraries/libutil/Sources/DefaultFilesystem.cpp
Ktwu: There's a memory filesystem I can try using for testing.

Ktwu:
> Is it possible to bump the minimum C++ version to C++17?

The spec for C++2017 was finalized last in December.  I don't know the usual upgrade cadence folks follow for new standards, but I suspect this is probably too new to define as the minimum.

PR #310, opened 2019-12-31
API Change, added new calls for AttributeList.h

NaruFGT:
AttributeList::Load( ... ) and AttributeList::write( ... ) can avoid taking the address of packed members within struct car_key_format *keyfmt.
Adjusted Sources/* to use these new functions.


NaruFGT:
It looks like Travis CI is having problems, I’m trying to reproduce the error, however my macOS is Catalina so I’m entirely sure the behaviour will be the same. I’m actually having problems with brew myself, installing gcc from brew hangs so I’m not to the point where Travis CI failed even.

PR #312, opened 2020-01-03
missing-prototypes

NaruFGT:
Contributor @strager had recommended enabling missing-prototypes, which proved useful for identifying unused code. In addition to missing-prototypes in C code, this will enable -Wall and -Werror.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,8 +61,19 @@ endif ()
 # Enable all warnings.
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" OR "${CMAKE_CXX_SIMULATE_ID}" STREQUAL "MSVC")
   add_compile_options(/D_CRT_SECURE_NO_WARNINGS) # TODO: /Wall /WX
-elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-  add_compile_options(-Wall -Werror)
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+        "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
+        "${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR
+        "${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
+  add_compile_options(-Wall -Werror -Wmissing-declarations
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -66,6 +66,8 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     add_compile_options(-Wno-multichar -Wno-sign-compare)
   endif ()
+elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
+  add_compile_options(-Wall -Werror -Wmissing-prototypes)
```
sas: I think this block will never be used. We already have an almost identical check at line 64. You wanna put `-Wmissing-prototype` at line 65.

sas (changes requested):

CMakeLists.txt
NaruFGT: The difference is whether we are invoking a C compiler or a C++ compiler. There are some libraries in this project which invoke the C compiler and CMake does not append C++ flags to invocations of the C compiler.
We cannot add -Wmissing-prototype to C++ CFLAGS because there are no "prototypes" in C++ and g++ will warn us which in our case with -Werror means those are errors.

NaruFGT: While it does need to distinguish whether this is a C compiler or a C++ compiler, I see that  add_compile_options() may not help with this discernment, I'm reviewing CMake documentation, I'll check to see if this is the correct approach.

NaruFGT: This is resolved with [this commit](https://github.com/facebook/xcbuild/pull/312/commits/32e37e9047c51deb25572ffd53331fab9bc83085).

sas (changes requested):
I don't think we should add these checks and then lower them to warning status unless we actually fix the issues in the code. The value provided by these warnings is low because it will just result in more noise during the build.

NaruFGT:
If we aren't going to address them then perhaps just disabling this warning entirely on GCC and Clang builds would be appropriate?

sas:
Disabling them completely would be fine by me. It's up to you; if you want we can disable them, but if you prefer fixing them and landing the fixes along with the changes that enable the warnings everywhere, then that works too.

Forgive me if I asked you this question before but are these warnings enabled by default on certain versions of gcc/clang? If I remember correctly, they are disabled by default and the user needs to enabled them.

